PR #11, opened 2020-10-21
Add Service Mesh interface

knkski:


charms/istio-pilot/reactive/charm.py
```diff
@@ -26,7 +26,9 @@ def configure_http(http):
     http.configure(hostname=hostname, port=hookenv.config()['xds-ca-tls-port'])
 
 
-@when_any("layer.docker-resource.oci-image.changed")
+@when_any(
+    "layer.docker-resource.oci-image.changed", "config.changed", "endpoint.service-mesh.changed"
```
kwmonroe, line 30: Hm... i'm a little nervous that you're acting on `ep.service-mesh.changed` because that flag [may not be cleared](https://github.com/juju-solutions/charms.reactive/blob/master/charms/reactive/endpoints.py#L53), meaning you'd clear the `started` flag on every hook.  Let's call in the big guns:  @johnsca does @knkski need to manually clear that `changed` flag in this handler?

johnsca: Yeah, agreed.  That flag won't be cleared automatically, so you should add an explicit clear for it.  May as well explicitly clear all of those flags, just in case.

johnsca (approved):

joedborg (approved):
LGTM

PR #14, opened 2020-10-28
Switch to whitelist for sidecar injections

knkski:
This avoids injecting sidecars into things that don't really need them


joedborg (approved):
LGTM

PR #22, opened 2021-03-19
Conversion to Operator Framework

DomFleischmann:
- Convert Charms to Operator Framework
- Add Unit Tests
- Add Linting
- Black Formatting


.github/workflows/linting-formatting.yaml
```diff
@@ -13,11 +13,19 @@ jobs:
...
         set -eux
         sudo apt-get install python3-setuptools
+        pip install tox
         pip3 install --user black==20.8b1
 
     - name: Check formatting
       run: ~/.local/bin/black --check .
+
+    - name: Linting
+      run: |
```
knkski, line 27: What do you think about keeping the linting at the top level, so we can just run one `tox -e lint` command?

.github/workflows/linting-formatting.yaml
```diff
@@ -13,11 +13,19 @@ jobs:
     - name: Check out repo
       uses: actions/checkout@v2
 
-    - name: Install black
+    - name: Install black and tox
       run: |
         set -eux
         sudo apt-get install python3-setuptools
+        pip install tox
```
knkski, line 20: Might be worth installing this via `apt` instead, since the version looks recent enough and we don't have to worry about `pip` vs `pip3`.

.github/workflows/unit-test.yaml
```diff
@@ -0,0 +1,28 @@
+name: Unit Tests 
+
+on:
+  - push
+  - pull_request
+
+jobs:
+  lint:
```
knkski, line 8: This should be updated to match the job name. Also, can we just coalesce this and `test-microk8s.yaml` into `ci.yaml` or similar? See here for where I went with `ci.yaml` and `cd.yaml`:

https://github.com/canonical/spark-operator/tree/master/.github/workflows

I'm not tied to that though, this would be a good opportunity to do some bikeshedding.

.github/workflows/unit-test.yaml
```diff
@@ -0,0 +1,28 @@
...
+    name: Run Unit Tests
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Check out repo
+      uses: actions/checkout@v2
+    - name: Setup Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.9
```
knkski, line 18: Do we actually need Python 3.9 specifically? Or can we just rely on the system version?

charms/istio-ingressgateway/requirements.txt
```diff
@@ -0,0 +1,4 @@
+ops
```
knkski, line 1: We should probably have version specifiers on these, if nothing other than `<x.0`, in case any of them publish breaking changes.

charms/istio-ingressgateway/src/charm.py
```diff
@@ -0,0 +1,229 @@
...
+
+from ops.charm import CharmBase
+from ops.main import main
+from ops.model import ActiveStatus, MaintenanceStatus, WaitingStatus, BlockedStatus
+
+from oci_image import OCIImageResource, OCIImageResourceError
+from k8s_service import RequireK8sService
+
+
+class IstioIngressGatewayCharm(CharmBase):
```
knkski, line 16: More bikeshedding, but can we switch all of these class names to just be `class Operator(CharmBase):`? I don't see much reason to name the class specific to the charm.

charms/istio-ingressgateway/src/charm.py
```diff
@@ -0,0 +1,229 @@
...
+        self.log = logging.getLogger(__name__)
+        self.istio_pilot = RequireK8sService(self, "istio-pilot")
+        self.image = OCIImageResource(self, "oci-image")
+        for event in [
+            self.on.install,
+            self.on.leader_elected,
+            self.on.upgrade_charm,
+            self.on.update_status,
+            self.on.config_changed,
+            self.istio_pilot.on.k8s_services_changed,
```
knkski, line 32: Is this a typo, or is `k8s-service` from metadata.yaml supposed to be pluralized here?

knkski: Also, is there any reason not to just use `self.on['istio-pilot'].relation_changed` here? I'm not sure what's going with `RequireK8sService` TBH. Is that an operator framework thing, or something we made?

charms/istio-ingressgateway/src/charm.py
```diff
@@ -0,0 +1,229 @@
...
+            image_details = self.image.fetch()
+        except OCIImageResourceError as e:
+            self.model.unit.status = e.status
+            self.log.info(e)
+            return
+
+        self.model.unit.status = MaintenanceStatus("Setting pod spec")
+
+        if not self.istio_pilot.is_created:
+            self.model.unit.status = BlockedStatus("Needing Istio Pilot relation")
```
knkski, line 47: Can this get written as maybe `Istio Pilot relation not found`? `Needing` seems a little vague.

charms/istio-ingressgateway/src/charm.py
```diff
@@ -0,0 +1,229 @@
...
+        if not self.istio_pilot.is_available:
+            self.model.unit.status = WaitingStatus("Waiting for Istio Pilot information")
+            return
+
+        pilot_service, pilot_port = self.istio_pilot.services[0]
+        cfg = self.model.config
+        namespace = self.model.name
+        service_name = self.model.app.name
+
+        config_map = self.check_ca_root_cert(namespace)
```
knkski, line 59: Can this get combined with the `if not self.istio_pilot.is_available` check above? It seems like the `Waiting for Istio Pilot information` status message is also appropriate for this check.

charms/istio-ingressgateway/src/charm.py
```diff
@@ -0,0 +1,229 @@
...
+                        ],
+                    }
+                ],
+            }
+        )
+
+        self.model.unit.status = ActiveStatus()
+
+    def check_ca_root_cert(self, namespace):
+        os.environ.update(
```
knkski, line 197: Can this get a comment with a link to https://bugs.launchpad.net/juju/+bug/1892255? It seems likely to make future code readers that are out of the loop go wtf.

charms/istio-pilot/src/charm.py
```diff
@@ -0,0 +1,340 @@
...
+                                    "failurePolicy": "Fail",
+                                    "namespaceSelector": {"matchLabels": {"juju-model": namespace}},
+                                    "objectSelector": {
+                                        "matchExpressions": [
+                                            {
+                                                'key': 'juju-app',
+                                                'operator': 'In',
+                                                'values': ['nonexistent']
+                                                # + [route['service'] for route in service_mesh.routes()],
+                                                + [route['service'] for route in []],
```
knkski, line 324: Is this supposed to be empty?

charms/istio-pilot/src/charm.py
```diff
@@ -0,0 +1,340 @@
...
+                                {
+                                    "name": "sidecar-injector.istio.io",
+                                    "clientConfig": {
+                                        "service": {
+                                            "name": service_name,
+                                            "namespace": namespace,
+                                            "path": "/inject",
+                                            "port": config.get('webhook-port'),
+                                        },
+                                        #  "caBundle": ca_bundle,
```
knkski, line 305: Is this supposed to be commented out?

.github/workflows/unit-test.yaml
DomFleischmann: I like unifying things, although I do think functional tests should be in a separate workflow. I like the idea of 3 workflows, linting & unit-tests, functional tests and publishing.

Regarding naming, I'm not entirely sure what we should use, I do think using acronyms like CI and CD sometimes ends up being a little confusing.

.github/workflows/unit-test.yaml
DomFleischmann: Correct, 3.9 is not necessary

charms/istio-ingressgateway/src/charm.py
DomFleischmann: This is an [interface](https://github.com/charmed-kubernetes/interface-k8s-service) we wrote some months ago for when charms need to connect to a service of another charm.

`k8s_services_changes` is a custom event of the interface which actually checks if the data has changed before sending out an event. It is pluralized because you can provide multiple services. You can checkout the code [here](https://github.com/charmed-kubernetes/interface-k8s-service/blob/main/k8s_service.py#L17).

charms/istio-pilot/src/charm.py
DomFleischmann: Good catch, forgot about that.

charms/istio-pilot/src/charm.py
DomFleischmann: This was already commented out in the reactive code, I'll just remove it, unless you remember the context of this remaining in there.

knkski: The context was probably that I forgot to remove it :smile:

knkski (approved):
:+1:

knkski:
@DomFleischmann: Can you squash these commits into one and and sign the resulting commit?

DomFleischmann:
> @DomFleischmann: Can you squash these commits into one and and sign the resulting commit?

Done!

PR #23, opened 2021-03-24
Install charmcraft in publish workflow

knkski:
With the switch to the operator framework, the publishing job needs to have charmcraft installed.


johnsca (approved):
Eventually, it would be good to move this to [actions-operator](https://github.com/charmed-kubernetes/actions-operator) and the tests to [pytest-operator](https://github.com/charmed-kubernetes/pytest-operator/) to handle this stuff for you, but this is good for now.

joedborg (approved):
LGTM

PR #25, opened 2021-04-19
Split up service-mesh interface

knkski:
Splits up service-mesh interface into ingress and ingress-auth interfaces, and uses the `schema` property for them. Punts on defining a full service-mesh interface until actually necessary to create.


DomFleischmann (commented):
Code LGTM but this test failure doesn't seem to be like the others

DomFleischmann (approved):

PR #29, opened 2021-06-02
Rejigger how tox works

knkski:
Running `tox` from the top-level now runs tox tests in each charm's directory, instead of a user having to cd to each subdir and run them individually.


DomFleischmann (approved):
LGTM

PR #31, opened 2021-08-25
Update Istio charms to 1.11

knkski:
Converts Istio charms to sidecar charms, and switches to using `istioctl` to spin up necessary resources.

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -1,23 +1,35 @@
 name: Continuous Integration
 
 on:
-  - push
-  - pull_request
+  push:
+    branches:
+      - master
+  pull_request:
 
 jobs:
-  unit-test:
+  lint:
+    name: Lint Code
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        charm: [pilot, gateway]
+    steps:
+    - uses: actions/checkout@v2
+    - run: sudo apt update && sudo apt install tox
+    - run: tox -e ${{ matrix.charm }}-lint
+
+  unit:
     name: Unit Test
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        charm: [pilot, ingressgateway]
-
+        charm: [pilot, gateway]
     steps:
     - uses: actions/checkout@v2
-    - run: sudo apt install tox
-    - run: tox -e ${{ matrix.charm }}
+    - run: sudo apt update && sudo apt install tox
+    - run: tox -e ${{ matrix.charm }}-unit
 
-  integration-test:
+  integration:
     name: Integration Test
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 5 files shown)


charms/istio-gateway/src/charm.py
```diff
@@ -0,0 +1,78 @@
...
+
+        self.log.info(rendered)
+        subprocess.run(["./kubectl", "apply", "-f-"], input=rendered.encode('utf-8'), check=True)
+
+        self.unit.status = ActiveStatus()
+
+    def remove(self, event):
+        """Remove charm."""
+
+        # Can't remove stuff yet: https://bugs.launchpad.net/juju/+bug/1941655
```
DomFleischmann, line 74: I think we have to talk a bit more about this bug, either we are doing something wrong or this is a pretty big issue for it being so low priority.

@jnsgruk what are your thoughts on this?

charms/istio-gateway/src/manifest.yaml
```diff
@@ -0,0 +1,324 @@
...
+                      - amd64
+                      - ppc64le
+                      - s390x
+      containers:
+        - args:
+            - proxy
+            - router
+            - --domain
+            - {{ namespace }}.svc.cluster.local
+            # - $(POD_NAMESPACE).svc.cluster.local
```
DomFleischmann, line 95: Can this be removed?

charms/istio-gateway/src/manifest.yaml
```diff
@@ -0,0 +1,324 @@
...
+              weight: 2
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+              - matchExpressions:
+                  - key: kubernetes.io/arch
+                    operator: In
+                    values:
+                      - amd64
+                      - ppc64le
+                      - s390x
```
DomFleischmann, line 88: Are we actuallly supporting these architectures?

charms/istio-gateway/src/manifest.yaml
```diff
@@ -0,0 +1,324 @@
...
+            # - $(POD_NAMESPACE).svc.cluster.local
+            - --proxyLogLevel=warning
+            - --proxyComponentLogLevel=misc:error
+            - --log_output_level=default:info
+          env:
+            - name: JWT_POLICY
+              value: third-party-jwt
+            - name: PILOT_CERT_PROVIDER
+              value: {{ pilot_host }}
+              # value: istiod
```
DomFleischmann, line 104: Is this still neeeded?

charms/istio-gateway/tests/test_charm.py
```diff
@@ -0,0 +1,20 @@
...
+
+def test_not_leader(harness):
+    harness.begin()
+    assert harness.charm.model.unit.status == WaitingStatus('Waiting for leadership')
+
+
+def test_basic(harness):
+    harness.set_leader(True)
+    harness.begin_with_initial_hooks()
+    assert harness.charm.model.unit.status == BlockedStatus('Config item `kind` must be set')
```
DomFleischmann, line 20: An additional test to make sure `kubectl` is called correctly would be nice here.

.github/workflows/integration.yaml
```diff
@@ -1,23 +1,35 @@
...
+        charm: [pilot, gateway]
     steps:
     - uses: actions/checkout@v2
     - run: sudo apt install tox
-    - run: tox -e ${{ matrix.charm }}
+    - run: tox -e ${{ matrix.charm }}-unit
 
-  integration-test:
+  integration:
     name: Integration Test
```
DomFleischmann, line 33: With this going to be a cross team effort I would like this to use pytest-operator.

charms/istio-gateway/src/manifest.yaml
knkski: Charmed Kubernetes supports those architectures to my knowledge, and is interested in using these charms for Ingress. Also, even if we stick with x86 only, I think it reduces a lot of maintenance burden to minimize diffs between upstream YAML and ours.

tests/test_bundle.py
```diff
@@ -0,0 +1,89 @@
...
+        'microk8s',
+        'kubectl',
+        'wait',
+        '--for=condition=available',
+        'deployment',
+        '--all',
+        '--all-namespaces',
+        '--timeout=5m',
+    )
+    await sleep(30)
```
johnsca, line 38: :nauseated_face:

PR #41, opened 2021-11-02
Better status handling

knkski:
Relies on these branches:

https://github.com/canonical/serialized-data-interface/tree/remove-object-inheritance
https://github.com/knkski/operator/tree/event-handler-returns

### DIFF
--- a/charms/istio-gateway/src/charm.py
+++ b/charms/istio-gateway/src/charm.py
@@ -6,29 +6,13 @@ import subprocess
 from jinja2 import Environment, FileSystemLoader
 from ops.charm import CharmBase
 from ops.main import main
-from ops.model import ActiveStatus, BlockedStatus, WaitingStatus
-from serialized_data_interface import NoCompatibleVersions, NoVersionsListed, get_interfaces
+from ops.model import BlockedStatus, WaitingStatus
+from serialized_data_interface import NoVersionsListed, get_interfaces
 
 
 class Operator(CharmBase):
     def __init__(self, *args):
         super().__init__(*args)
-
-        if not self.unit.is_leader():
-            # We can't do anything useful when not the leader, so do nothing.
-            self.model.unit.status = WaitingStatus("Waiting for leadership")
-            return
-
-        try:
-            self.interfaces = get_interfaces(self)
-        except NoVersionsListed as err:
-            self.model.unit.status = WaitingStatus(str(err))
-            return
-        except NoCompatibleVersions as err:
-            self.model.unit.status = BlockedStatus(str(err))
-            return
-        else:
-            self.model.unit.status = ActiveStatus()
 
         self.log = logging.getLogger(__name__)
 
@@ -40,17 +24,22 @@ class Operator(CharmBase):
     def install(self, event):
         """Install charm."""
 
+        if not self.unit.is_leader():
+            return self.unit, WaitingStatus("Waiting for leadership")
+
+        try:
+            self.interfaces = get_interfaces(self)
+        except NoVersionsListed as err:
+            return self.app, WaitingStatus(str(err))
+
         if self.model.config['kind'] not in ('ingress', 'egress'):
-            self.model.unit.status = BlockedStatus('Config item `kind` must be set')
-            return
+            return self.app, BlockedStatus('Config item `kind` must be set')
 
         if not self.model.relations['istio-pilot']:
-            self.model.unit.status = BlockedStatus("Waiting for istio-pilot relation")
-            return
+            return self.app, BlockedStatus("Waiting for istio-pilot relation")
 
         if not ((pilot := self.interfaces["istio-pilot"]) and pilot.get_data()):
-            self.model.unit.status = WaitingStatus("Waiting for istio-pilot relation data")
-            return
+            return self.app, BlockedStatus("Waiting for istio-pilot relation data")
 
         pilot = list(pilot.get_data().values())[0]
 
@@ -64,8 +53,6 @@ class Operator(CharmBase):
         )
 
         subprocess.run(["./kubectl", "apply", "-f-"], input=rendered.encode('utf-8'), check=True)
-
-        self.unit.status = ActiveStatus()
 
     def remove(self, event):
         """Remove charm."""
--- a/charms/istio-gateway/requirements.txt
+++ b/charms/istio-gateway/requirements.txt
@@ -1,4 +1,6 @@
 jinja2<3.1
-ops<1.3.0
+# ops<1.3.0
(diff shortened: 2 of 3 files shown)


DomFleischmann (commented):
I'm all in favor of moving the leadership and interface checks to the `install` method. 

But I'm still not convinced about storing the BlockedStatus in stored state. Looking at the [ops changes](https://github.com/knkski/operator/commit/d36a21375fa5158b6af0b1dca1f6e1a0b6880787#)  I'm not confident that the status will be cleaned up correctly. Secondly stored state is on unit level which I'm not entirely sure if it is taken into account. 

All in all I'm very opposed to handling any kind of status with stored state and I believe this will not get good reception from the operator framwork team or the juju team either. I think our main issue should be almost solved with moving the status handling to the install method and maybe considering `event.defer()` in the case we are unsure a status will be set again.

charms/istio-gateway/src/charm.py
```diff
@@ -40,17 +24,22 @@ def __init__(self, *args):
...
+
+        try:
+            self.interfaces = get_interfaces(self)
+        except NoVersionsListed as err:
+            return self.app, WaitingStatus(str(err))
+
         if self.model.config['kind'] not in ('ingress', 'egress'):
-            self.model.unit.status = BlockedStatus('Config item `kind` must be set')
-            return
+            return self.app, BlockedStatus('Config item `kind` must be set')
```
ca-scribner, line 36: I'm not sure returning status on hooks is rich enough.  For example, what if an `update_status` hook checks relations and finds two are broken - how does that get surfaced?

DomFleischmann:
Let's abandon this PR and focus on improving [this](https://github.com/canonical/istio-operators/pull/40#pullrequestreview-796247060) one. After talking with @jnsgruk we don't really see a way of getting these kind of changes merged into the operator framework.

knkski:
@DomFleischmann one nice thing about this method is that it doesn't matter if using StoredState is sub-optimal, as that's an implementation detail that is handled in the operator framework. I think the operator framework should juggle how to display statuses like these to the user, as that will let the operator framework be updated to work nicely with future juju features such as rich statuses, or relation-level statuses.

I'm also not sure what you dislike about handling statuses with `StoredState`. As far as I can tell, that's what `StoredState` is there for.

knkski:
@jnsgruk can you comment on what you think would block something like this from making its way into the operator framework? It really seems like the operator framework itself should handle displaying rich statuses like this to the user, particularly because that will work nicely with Juju's concept of rich statuses and/or relation-level statuses when those get implemented.

charms/istio-gateway/src/charm.py
knkski: I don't think `update_status` hooks should really be used at all, as they're semi-broken. However if you really wanted to use one like that, you could just do `return self.app, BlockedStatus("There are two things broken: foo, bar")`. The handling over in the operator framework would still take care of updating the status when `update_status` hook ran again, and did ``return self.app, BlockedStatus("There is one thing broken: foo")` instead.

Also, one advantage of this method is if it doesn't work for a particular use case, a charm author can `return None` and continue to manually set `self.model.unit.status`, as they can currently do.

DomFleischmann:
@knkski I am not a big fan of `StoredState` in general, one good thing about an event driven framework is that you can treat events in an isolated fashion, `StoredState` contradicts that. In the case of handling Application/Unit Status I think it is a bad idea trusting the state of previous events, especially when checking the current situation of the entity does not cost much.

On the second part I am also against the operator framework setting any kind of Status. Application/Unit Status is something that the charmers should decide on as they are the ones that know the application best. If suddenly status changes occur due to something that happens in the operator framework it will be hard to debug, especially for new charmers.

Regarding how operator framework will deal with future features like rich statuses, we don't even know how that feature will look like, so I don't think we are making the operator framework team a lot of favors with our assumptions.

knkski:
@DomFleischmann I am also not a fan of `StoredState` in general, but it's an implementation detail that we don't/shouldn't have to care about. When you say:

> I think it is a bad idea trusting the state of previous events, especially when checking the current situation of the entity does not cost much.

I'm not sure how this is different from the way things work today. If a Juju hook runs and an event handler sets a status, until that event handler runs again, the status (i.e. `juju status`) will depend on the state of that previous event.

I also am not sure what you mean by `I am also against the operator framework setting any kind of Status`. All of the statuses are getting set explicitly because a charm author wanted that status to be set. There's never a "suddenly status changes occur due to something that happens in the operator framework" situation?

ca-scribner:
@DomFleischmann do you have examples of `StoredState` going wrong?  I'm not sure I prefer it, but I'm having trouble thinking of a situation where it breaks us in this situation.  It is unit-only, but so are the statuses we're setting.  Is the case where it breaks us this?

1. `relation_hook` puts `BlockedStatus` to `StoredState`
2. `unit` is deleted (eg: `pod` restarted)
3.  new unit is created.  **what hooks fire in this case?**
4. `update_status` fires and rather than check relations, it relies on past `relation_hook` data in `StoredState`

I think the part that will make this work/break are what hooks fire/what code is run in step (3).  Does anyone know?

ca-scribner:
@knkski regarding `operator framework setting` from @DomFleischmann what I think he means is what is happening [here](https://github.com/knkski/operator/blob/d36a21375fa5158b6af0b1dca1f6e1a0b6880787/ops/framework.py#L809), where the framework is modifying the status (based on data we've previously given it, but still).

knkski:
@DomFleischmann I think what you are arguing for is some sort of `juju get-status my-charm` functionality, where `juju status` goes off and fetches the current state of things instead of relying on status from past events. I think that would be interesting, but also a much bigger ask from both the operator framework and from Juju than what is proposed here.

DomFleischmann:
@ca-scribner @knkski What I'm arguing for is that we should handle status in our charm code like this [PR](https://github.com/canonical/istio-operators/pull/40) is doing. We got ourselves into this situation by handling status wrong, I'm 100% sure we don't need a new feature in juju or in the operator framework to get ourselves out of it. Let's focus on the fact that we want to get a release out and our charms are currently showing the wrong status.

I'm happy to discuss further all the objections that I have with this approach of handling status in the operator framework. But I don't feel like it is getting us any closer to fixing our real issue. If we want to have features like this in operator framework or juju let's involve the people that are responsible for these projects. If you want to discuss this PR further let's schedule a meeting and write down our conclusions afterwards here. But none of these features are required to fix our real issue, so let's spend our energy on improving the [PR](https://github.com/canonical/istio-operators/pull/40) that allow us to fix it on our own.

PR #43, opened 2021-12-01
Add support for per-unit routing

johnsca:
Due to some unfortunate technical limitations / requirements, there are use-cases which require ingress to individual units in addition to the application service as a whole.

This also adds the ingress URL(s) to the response data for the requesting charm.

Depends on: https://github.com/canonical/operator-schemas/pull/4


charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
```diff
@@ -0,0 +1,189 @@
...
+class IngressRequirerEvents(ObjectEvents):
+    available = EventSource(IngressProviderAvailableEvent)
+    ready = EventSource(IngressReadyEvent)
+    failed = EventSource(IngressFailedEvent)
+
+
+class IngressRequirer(Object):
+    on = IngressRequirerEvents()
+
+    def __init__(self, charm: CharmBase, relation_endpoint: str):
```
mmanciop, line 49: I think we should use the same pattern with the default relation name that has proven to be very successful for the O11y team: https://github.com/canonical/prometheus-k8s-operator/blob/37a82abd1909e59aa5ac0ff810cca74864ca13d4/lib/charms/prometheus_k8s/v0/prometheus_scrape.py#L835

mmanciop: Also, notice the logic to validate the role of the relation, it saves lives: https://github.com/canonical/prometheus-k8s-operator/blob/37a82abd1909e59aa5ac0ff810cca74864ca13d4/lib/charms/prometheus_k8s/v0/prometheus_scrape.py#L857

mmanciop: Last but not list, I cannot imagine _any scenario_ in which a requiring charm may want a relation with `limit` value different than `1` (why would you connect to _multiple ingress controllers_ at once?). We should probably enforce that.

charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
```diff
@@ -0,0 +1,189 @@
...
+            per_unit_routes: whether or not to create URLs which map to specific units;
+                the URLs will have their own prefix of f"{prefix}-unit-{unit_num}" (with
+                tailing slashes handled appropriately)
+        """
+        if not self.charm.unit.is_leader():
+            raise RequestFailed(
+                WaitingStatus(f"Only leader can request ingress: {self.relation_endpoint}")
+            )
+        if self.status is not None:
+            raise RequestFailed(self.status)
```
mmanciop, line 131: I actually would expect the requirer charm to specify these parameters in the constructor of `IngressRequirer`. That tends to make the usage of the library more resilient, as you do not need to hook it specifically anywhere in your charm, just instantiate `IngressRequirer` in your charm's constructor.

charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
```diff
@@ -0,0 +1,189 @@
...
+        all_data = ingress.get_data()
+        for (rel, app), data in all_data.items():
+            if app is self.charm.app:
+                continue
+            return data["url"]
+        else:
+            return None
+
+    @cached_property
+    def unit_urls(self):
```
mmanciop, line 165: I would raise if per unit routing is not configured.

charms/istio-pilot/metadata.yaml
```diff
@@ -22,8 +22,9 @@ provides:
     versions: [v1]
   ingress:
     interface: ingress
-    schema: https://raw.githubusercontent.com/canonical/operator-schemas/master/ingress.yaml
-    versions: [v1, v2]
+    #schema: https://raw.githubusercontent.com/canonical/operator-schemas/master/ingress.yaml
+    schema: https://raw.githubusercontent.com/canonical/operator-schemas/1ed74c640bc289b71f261cda67177ee5209a1562/ingress.yaml
+    versions: [v1, v2, v3]
   ingress-auth:
```
mmanciop, line 28: What does this do? I cannot figure it out from the code :D

charms/istio-pilot/src/charm.py
```diff
@@ -181,23 +239,55 @@ def handle_ingress_auth(self, event):
...
-        subprocess.run(
+        self._kubectl(
+            'delete',
+            'envoyfilters,rbacconfigs',
+            f'-lapp.juju.is/created-by={self.app.name}',
+        )
+
+        self._kubectl("apply", "-f-", input=manifests)
+
+    def _kubectl(self, *args, namespace=None, input=None, capture_output=False):
```
mmanciop, line 250: `lightkube` works like a charm (eh!), see https://github.com/canonical/prometheus-k8s-operator/blob/main/lib/charms/observability_libs/v0/kubernetes_service_patch.py

mmanciop (changes requested):
Have not tried to integrate it yet, but I am hopeful :D

I think there are some improvements to the ergonomics of the library that should be considered, see inline.

charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
johnsca: Apologies; I don't understand this comment. Creating the instance of the `IngressRequirer` in the charm's `__init__` is exactly the pattern being used.  I also don't know what "parameters" you're referring to here, as this is just error checking and propagating an apparently ignored bad status as an exception with the same info.

mmanciop: So, what the code seems to imply is that you will have to invoke the `request` method somewhere in a hook. I am saying that the same parameters you pass to `request` you could rather be passing to the `IngressRequirer` constructor (https://github.com/canonical/istio-operators/pull/43/files#diff-e846165da73fe4674f0e581c3df778471bec5f1c5caab26f06437772fd4eeda8R49), and do not expect the charm to invoke any method to "trigger" the requesting of an ingress mapping: that would happen automatically when the relation between requirer charm and the ingress charm is established.

johnsca: That would restrict a give charm to requesting ingress to only a single service per relation to the ingress provider. That seems like an unnecessary arbitrary restriction, given that there's no reason a particular operator charm can't deploy multiple services or even define services dynamically based on things like other relations or config. (Granted, the way the underlying schema is currently defined also imposes this limitation, but now that you've called that out, I think I should fix the new schema PR rather than changing how this part works).

charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
johnsca: That seems unnecessarily aggressive.  If you didn't request per-unit routes, you won't get any, and I don't really see that as an error.  Additionally, checking that would require reading the app-level relation data sent by this application which is restricted to only the leader by Juju which would mean non-leader units could never read their own URL. (There is [an open bug against SDI](https://github.com/canonical/serialized-data-interface/issues/27) which also imposes this limitation, but I intend to submit a fix for that.)

charms/istio-pilot/metadata.yaml
johnsca: The `ingress` interface and the istio-pilot operator charm were designed using the [Serialized Data Interface (SDI)](https://github.com/canonical/serialized-data-interface) library, and it uses these non-canonical fields to reference the schema and note which versions are supported by the charm.

charms/istio-pilot/src/charm.py
johnsca: Yep!  There is a separate set of work going on to move this over to lightkube, but it is being done outside of this feature PR.

charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
johnsca: > why would you connect to multiple ingress controllers at once?

Well, actually, I could see a possible use-case where you want to SaaS a service to multiple separate networks and want to maximize isolation between those networks so you have multiple ingress controllers, each of which only has a network connection to one of the external networks. It's a pretty out-there example, but I don't know that it's worth putting in extra effort to add in an artificial limitation. Then again, knowing which ingress address to advertise to a particular related app would run up against the issue that Vault did where Juju doesn't provide enough info about a relation to know which address to present to it and the charm has to sort of guess based on things like subnet overlap. *shrug*

charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
mmanciop: Writing is restricted to the leader, not reading :-)

charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
mmanciop: Actually, I think your good intentions will result in a confusing UX for charm authors. In the case of wanting to expose independently different parts of a charm’s workload API, it makes for a better DevX (and also admin UX), to have separate relations using the `ingress` relation interface. That is, in the prometheus-k8s charm, we would declare two relations with the `ingress` relation interface, named, for example, `ingress-remote-write` and `ingress-web-ui`.

The alternative is… what? Having the charm author put boolean switches in their charm’s config to give the Juju admin the option of deciding which APis to expose? That would be super un-Juju :-)

The model I propose also means that your code can remain super simple :-)

charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
mmanciop: But then how do you map those unit/app URIs to what you put in which configs. It sounds theoretically possible, but I do not see it as a problem worth solving atm :-)

johnsca: It has, in fact, come up with Vault when they needed different addresses sent to different client relations based on what network segment they are in (internal vs external), so it's not theoretical in the slightest. However, sorting that out properly requires some changes to how Juju presents info about a given relation to the charm, so it's currently only partially supported in Vault today (they use a heuristic based on subnets, which fails in many environments).

My point was just that it would take additional effort to try to enforce the limit (esp. when Juju already does that if you put it in), particularly given https://github.com/canonical/operator/issues/498, so I'm disinclined to put effort into that right now.

mmanciop: The reason why we tend to enforce these limits in the logic of the library is that charm authors seldom think to think that aspect through. The default behavior of `limit` being "Pile on as many relations instances as you please" in my experience plays absolute havoc on the mental model of charm authors that predominantly seem to assume that "I have _up to one_ relation of this kind".

charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
johnsca: Supporting the request / response pattern does not preclude the charm author from defining multiple relation endpoints, and we did just that in the CK charms for the `loadbalancer` interface because it made sense in that context.  But there could well be other contexts where it would make more sense for a charm to expose a group of services with the same ingress characteristics.

Also, this fundamentally *is* a request / response pattern, because the properties of the ingress request must be sent over once the version negotiation is successful, before the response can be returned. Hiding that in the constructor doesn't change that fact and just obscures the actual flow of logic for minimal or no gain in usability. (Whether the params are passed in to the constructor or a method, they still have to be passed in. And doing so in a handler gives the charm author the chance to set an appropriate status, should they want, which they can't do under your proposal.)

All that said, this is not a hill I'm willing to die on, because I'm still just trying to get this thing working and useful at this point. :sweat_smile:

charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
johnsca: Reading your own app-level relation data is, in fact, restricted to the leader by Juju. I don't agree with that, and maybe it will change at some point, but AFAIK that is still a limitation.

johnsca: I was a little unclear about the SDI bug; that *also* prevents non-leaders from reading *remote* app-level relation data, which is, in fact, allowed by Juju.

charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
johnsca: It would be nice if the library could have a build-time hook to validate the `limit` field before the charm is even built.

johnsca: Added all of this.

charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
johnsca: Added optional request init param support so it can be used with this pattern, or with an explicit call to `request`.

charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
```diff
@@ -0,0 +1,268 @@
...
+            "service": service,
+            "prefix": prefix,
+            "rewrite": rewrite,
+            "namespace": namespace,
+            "per_unit_routes": per_unit_routes,
+        }
+        if any(self._request_args.values()) and not self._request_args["port"]:
+            raise TypeError("Missing required argument: 'port'")
+
+        if charm.unit.is_leader():
```
mmanciop, line 113: This breaks access to per-unit routing info to non-leader units

charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
```diff
@@ -0,0 +1,268 @@
...
+            self.framework.observe(charm.on[relation_name].relation_created, self._check_provider)
+            self.framework.observe(charm.on[relation_name].relation_changed, self._check_provider)
+            self.framework.observe(charm.on[relation_name].relation_broken, self._lost_provider)
+            self.framework.observe(charm.on.leader_elected, self._check_provider)
+
+    def _get_status(self):
+        if not self.charm.unit.is_leader():
+            return None
+        if not self.charm.model.relations[self.relation_name]:
```
mmanciop, line 122: This also breaks access to per-unit routing info to non-leader units

mmanciop (changes requested):

charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
```diff
@@ -0,0 +1,268 @@
...
+
+class IngressRequirer(Object):
+    on = IngressRequirerEvents()
+
+    def __init__(
+        self,
+        charm: CharmBase,
+        relation_name: str = DEFAULT_RELATION_NAME,
+        *,
+        port: int = None,
```
mmanciop, line 62: I feel that, if unspecified, we should expose all the ports in the service

charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
johnsca: Yes, but it's currently required to work around https://github.com/canonical/serialized-data-interface/issues/27. Doesn't the new prometheus interface also use app-level relation data, in which case only the leader need be involved anyway?

charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
johnsca: How?  Non-leaders can't send an ingress request, so they shouldn't need to care about the status.

charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
johnsca: There's no way to know that except by introspecting the K8s service directly, which goes against how relations are intended to work in Juju.  Besides, Istio only accepts a single port in the [`Destination`](https://istio.io/latest/docs/reference/config/networking/virtual-service/#Destination) specification, anyway (and requires it whenever a service does expose multiple ports, to disambiguate).

johnsca: To clarify on that last bit, Istio is mapping a single HTTP gateway + port to multiple backend services via paths. There's no way to do that for multiple ports unless you request a separate mapping (i.e. path) for each desired port. If multiple ports need to be exposed, this would be a use-case for a charm sending multiple ingress requests on a single relation.

mmanciop: > There's no way to know that except by introspecting the K8s service directly, which goes against how relations are intended to work in Juju

I am failing to understand the "goes against how relations are intended to work in Juju". I do not see anything un-Juju in there tbh.

> Istio only accepts a single port in the Destination specification

OK, so if unspecified, let's raise if the service has more than one port :-)

charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
mmanciop: > Non-leaders can't send an ingress request, so they shouldn't need to care about the status

Non-leaders are supposed to be able to follow the events, right? (At least, Prometheus does :-) ) Ingress events are fired only in the leader, though. So I do not think this limitation is acceptable in per-unit routing.

charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
mmanciop: Nope. For Prometheus, each unit is on its own.

charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
johnsca: > I am failing to understand the "goes against how relations are intended to work in Juju". I do not see anything un-Juju in there tbh.

I mean that in theory the interface could be used outside of K8s for some other type of ingress provider / consumer, in which case there would be no way to "introspect the service" to determine the default port. Although, TBH, the `service` field already breaks that a bit; it should really be `destination` and go back to being required, since whether it's a service name (which is really a host name in K8s) or the ingress address would depend on what type of charm is using the relation.

mmanciop: > I mean that in theory the interface could be used outside of K8s for some other type of ingress provider / consumer

In that case I would expect a library with a similar or same API, owned by a different charm.

charms/istio-pilot/src/charm.py
```diff
@@ -95,17 +124,36 @@ def send_info(self, event):
...
+                if app != self.app
+            }
         else:
-            routes = []
+            routes = {}
+
+        if isinstance(event, RelationBrokenEvent):
+            # The app-level data is still visible on a broken relation, but we
+            # shouldn't be keeping the VirtualService for that related app.
+            del routes[(event.relation, event.app)]
```
ca-scribner, line 151: Has this been tested?  I cherry picked this in #49 but when testing it encountered this error when I broke a relation:
```
  File "./src/charm.py", line 266, in <module>
    main(Operator)
  File "/var/lib/juju/agents/unit-istio-pilot-0/charm/venv/ops/main.py", line 406, in main
    _emit_charm_event(charm, dispatcher.event_name)
  File "/var/lib/juju/agents/unit-istio-pilot-0/charm/venv/ops/main.py", line 140, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-istio-pilot-0/charm/venv/ops/framework.py", line 278, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-istio-pilot-0/charm/venv/ops/framework.py", line 722, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-istio-pilot-0/charm/venv/ops/framework.py", line 767, in _reemit
    custom_handler(event)
  File "./src/charm.py", line 143, in handle_ingress
    del routes[(event.relation, event.app)]
TypeError: 'dict_items' object does not support item deletion
unit-istio-pilot-0: 11:10:39 ERROR juju.worker.uniter.operation hook "ingress-relation-broken" (via hook dispatching script: dispatch) failed: exit status 1
unit-istio-pilot-0: 11:10:39 INFO juju.worker.uniter awaiting error resolution for "relation-broken" hook
```

I may have broken something myself in the process though.  I'm going to add a unit test and debug this in #49

ca-scribner: nevermind, I did not also cherry pick the updated way routes was constructed (right above this), so I had a dict_items object instead of a dict. I think this was my own mistake and yours works (although a relation-broken test would still help)

johnsca: Yeah, that looks like a cherry-pick issue.  The relation remove is covered by [the cleanup portion](https://github.com/canonical/istio-operators/pull/43/files#diff-f602add18718f057def3bdfbfb4f0f923ae6cfe123a82b2ff8f639de063fe288R166-R167), and that error implies that you must be missing [this bit of the change](https://github.com/canonical/istio-operators/pull/43/files#diff-d3a4e684bd1a3a1822df5ab7605cfce00c97cd7b52f44ed770526c4f4f80eacdR136-R146) which makes `routes` an actual `dict` rather than the `dict_items` list-like.

charms/istio-pilot/src/charm.py
```diff
@@ -39,10 +39,13 @@ def __init__(self, *args):
         self.framework.observe(self.on.install, self.install)
         self.framework.observe(self.on.remove, self.remove)
 
+        self.framework.observe(self.on.config_changed, self.handle_default_gateways)
+
         self.framework.observe(self.on["istio-pilot"].relation_changed, self.send_info)
 
         self.framework.observe(self.on['ingress'].relation_changed, self.handle_ingress)
         self.framework.observe(self.on['ingress'].relation_departed, self.handle_ingress)
+        self.framework.observe(self.on['ingress'].relation_broken, self.handle_ingress)
```
ca-scribner, line 48: Should we be subscribed to both of these events?  Departed happens if a unit leaves a relation, and broken happens if all units leave the relation (and thus the relation is removed).  Do both have significance for us, or do we need only `relation_broken`?  I don't think we do anything different depending on the number of units on the other side

If both matter, should [this line](https://github.com/canonical/istio-operators/pull/43/files#diff-d3a4e684bd1a3a1822df5ab7605cfce00c97cd7b52f44ed770526c4f4f80eacdR148) catch `RelationDepartedEvent` as well?  I think no, as we don't want to delete the virtual service just because some (but not all) units of our related app have left, but thought I'd mention it just in case

johnsca: Departed only matters if per-unit routing is being used, but in that case it does matter, yes, since we want to remove that route from the VS (but not the entire VS).

charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
```diff
@@ -0,0 +1,305 @@
...
+LIBAPI = 0
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 1
+
+
+SCHEMA_URL_BASE = "https://raw.githubusercontent.com/canonical/operator-schemas"
+# SCHEMA_URL = f"{SCHEMA_URL_BASE}/master/ingress.yaml"
+SCHEMA_URL = f"{SCHEMA_URL_BASE}/1ed74c640bc289b71f261cda67177ee5209a1562/ingress.yaml"
```
johnsca, line 38: TODO: Remove once https://github.com/canonical/operator-schemas/pull/4 is merged and before this is.

charms/istio-pilot/metadata.yaml
```diff
@@ -22,8 +22,9 @@ provides:
     versions: [v1]
   ingress:
     interface: ingress
-    schema: https://raw.githubusercontent.com/canonical/operator-schemas/master/ingress.yaml
-    versions: [v1, v2]
+    #schema: https://raw.githubusercontent.com/canonical/operator-schemas/master/ingress.yaml
+    schema: https://raw.githubusercontent.com/canonical/operator-schemas/1ed74c640bc289b71f261cda67177ee5209a1562/ingress.yaml
```
johnsca, line 26: TODO: Remove once https://github.com/canonical/operator-schemas/pull/4 is merged and before this is.

johnsca (commented):
Marking some TODOs to ensure they're done before this is merged.

charms/istio-pilot/src/charm.py
```diff
@@ -115,36 +163,48 @@ def get_kwargs(version, route):
...
-            check=True,
+        self._kubectl(
+            'delete',
+            'virtualservices,destinationrules',
+            f'-lapp.juju.is/created-by={self.app.name}',
         )
+        if routes:
+            self._kubectl("apply", "-f-", input=virtual_services)
+
+        # Send URL(s) back
```
knkski, line 192: What happens if a cluster exposes multiple ingresses for the same set of services, e.g. `kubeflow.company.com` and `kubeflow.company.vpn`/`kubeflow-vpn.company.com`? Or a situation like `old.reddit.com`/`www.reddit.com`/`my-subreddit.reddit.com`, where the prefix has meaning, but is handled by the same set of servers.

It really doesn't seem like `ingress` should be handling any notion of public URLs. As it is currently, an `ingress` requirer says to the provider "I'd like to be routed to, here's how to talk to me", and doesn't care which public URL is used, as that's up to how the provider is configured.

Introducing the notion of public URL into `ingress` forces charmed services to know/care about things that they really shouldn't.

mmanciop:
mmm more bugs: 

```
============================================================ FAILURES =============================================================
_______________________________________________ TestCharm.test_ingress_relation_set _______________________________________________
  File "/home/michele/git/prometheus-operator/tests/unit/test_charm.py", line 103, in test_ingress_relation_set
    self.assertEqual(
  File "/usr/lib/python3.9/unittest/case.py", line 831, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/usr/lib/python3.9/unittest/case.py", line 824, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: None != 'http://prometheus-k8s:9090'
_____________________________ TestRemoteWriteProvider.test_endpoint_url_with_ingress_and_external_url _____________________________
  File "/home/michele/git/prometheus-operator/tests/unit/test_remote_write.py", line 186, in test_endpoint_url_with_ingress_and_external_url
    self.harness.begin_with_initial_hooks()
  File "/home/michele/git/prometheus-operator/.tox/unit/lib/python3.9/site-packages/ops/testing.py", line 215, in begin_with_initial_hooks
    self._charm.on.config_changed.emit()
  File "/home/michele/git/prometheus-operator/.tox/unit/lib/python3.9/site-packages/ops/framework.py", line 276, in emit
    framework._emit(event)
  File "/home/michele/git/prometheus-operator/.tox/unit/lib/python3.9/site-packages/ops/framework.py", line 736, in _emit
    self._reemit(event_path)
  File "/home/michele/git/prometheus-operator/.tox/unit/lib/python3.9/site-packages/ops/framework.py", line 783, in _reemit
    custom_handler(event)
  File "/home/michele/git/prometheus-operator/src/charm.py", line 129, in _configure
    new_layer = self._prometheus_layer
  File "/home/michele/git/prometheus-operator/src/charm.py", line 328, in _prometheus_layer
    "command": self._command(),
  File "/home/michele/git/prometheus-operator/src/charm.py", line 194, in _command
    if unit_external_url := self.ingress.unit_url:
  File "/usr/lib/python3.9/functools.py", line 969, in __get__
    val = self.func(instance)
  File "/home/michele/git/prometheus-operator/lib/charms/istio_pilot/v0/ingress.py", line 295, in unit_url
    return self.unit_urls[self.charm.unit.name]
TypeError: 'NoneType' object is not subscriptable
```

Due to the following being not `None`-safe:

```
    @cached_property
    def unit_url(self):
        """The full ingress URL which map to the current unit.

        May return None if the URLs aren't available yet, or if per-unit routing
        was not requested. Otherwise, returns a URL string.
        """
        return self.unit_urls[self.charm.unit.name]
```

mmanciop:
This happens when I remote the relation between a charm using the `ingress` library and `istio-pilot`:

```
unit-prometheus-0: 14:09:46 ERROR unit.prometheus/0.juju-log ingress:7: Uncaught exception while in charm code:
  File "/var/lib/juju/agents/unit-prometheus-0/charm/venv/ops/model.py", line 1521, in _run
    result = run(args, **kwargs)
  File "/usr/lib/python3.8/subprocess.py", line 512, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '('/var/lib/juju/tools/unit-prometheus-0/relation-get', '-r', '7', '-', '', '--app', '--format=json')' returned non-zero exit status 1.

During handling of the above exception, another exception occurred:

  File "./src/charm.py", line 347, in <module>
    main(PrometheusCharm)
  File "/var/lib/juju/agents/unit-prometheus-0/charm/venv/ops/main.py", line 414, in main
    charm = charm_class(framework)
  File "./src/charm.py", line 73, in __init__
    self.ingress = IngressRequirer(
  File "/var/lib/juju/agents/unit-prometheus-0/charm/lib/charms/istio_pilot/v0/ingress.py", line 113, in __init__
    self.status = self._get_status()
  File "/var/lib/juju/agents/unit-prometheus-0/charm/lib/charms/istio_pilot/v0/ingress.py", line 136, in _get_status
    self._get_interface()
  File "/var/lib/juju/agents/unit-prometheus-0/charm/lib/charms/istio_pilot/v0/ingress.py", line 160, in _get_interface
    return SerializedDataInterface(
  File "/var/lib/juju/agents/unit-prometheus-0/charm/venv/serialized_data_interface/__init__.py", line 110, in __init__
    others = {
  File "/var/lib/juju/agents/unit-prometheus-0/charm/venv/serialized_data_interface/__init__.py", line 111, in <dictcomp>
    app.name: bag.get("_supported_versions")
  File "/usr/lib/python3.8/_collections_abc.py", line 660, in get
    return self[key]
  File "/var/lib/juju/agents/unit-prometheus-0/charm/venv/ops/model.py", line 430, in __getitem__
    return self._data[key]
  File "/var/lib/juju/agents/unit-prometheus-0/charm/venv/ops/model.py", line 414, in _data
    data = self._lazy_data = self._load()
  File "/var/lib/juju/agents/unit-prometheus-0/charm/venv/ops/model.py", line 779, in _load
    return self._backend.relation_get(self.relation.id, self._entity.name, self._is_app)
  File "/var/lib/juju/agents/unit-prometheus-0/charm/venv/ops/model.py", line 1588, in relation_get
    return self._run(*args, return_output=True, use_json=True)
  File "/var/lib/juju/agents/unit-prometheus-0/charm/venv/ops/model.py", line 1523, in _run
    raise ModelError(e.stderr)
ops.model.ModelError: b'ERROR "" is not a valid unit or application\n'
```

100% reproducible.

mmanciop (changes requested):
See bugs reported as comments on the PR.

mmanciop:
Also, how would I be able to unit test the library integration? Specifically, how do I mock the relation data sent by the ingress to a unit using the library?

charms/istio-pilot/src/charm.py
johnsca: Agreed.  The relation is just "tell me what my ingress address is" and whether that's public with an external hostname or just across internal clusters, that's up to the ingress provider (Istio, in this case) to manage.  Typically, that would be done by config on either whatever is providing the `LoadBalancer` services their external address, or config on the provider charm to override it with something that's managed entirely externally (like a public DNS entry).

charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
```diff
@@ -0,0 +1,305 @@
...
+                "namespace": namespace or self.model.name,
+                "prefix": prefix or f"/{self.charm.app.name}/",
+                "rewrite": rewrite or "/",
+                "service": service or self.charm.app.name,
+                "port": port,
+                "per_unit_routes": per_unit_routes,
+            },
+        )
+
+    @cached_property
```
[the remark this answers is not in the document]
johnsca: It's only cached in memory, and the value won't change within a given hook execution, so no.

charms/istio-pilot/lib/charms/istio_pilot/v0/ingress.py
```diff
@@ -0,0 +1,305 @@
...
+        still defers to what's in the metadata.yaml, however, so that local_sdi.py
+        can be used to inline the schema into the charm to avoid runtime network
+        access, if desired.
+        """
+        # Can't use self.charm.meta, unfortunately, because it doesn't preserve
+        # the schema or versions fields.
+        meta = yaml.safe_load(Path("metadata.yaml").read_text())
+        endpoint_spec = meta["requires"][self.relation_name]
+        schema = get_schema(endpoint_spec.get("schema", SCHEMA_URL))
+        versions = endpoint_spec.get("versions", SCHEMA_VERSIONS)
```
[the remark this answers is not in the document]
johnsca: The wrapper library's code is inherently tied to the specific schema versions it supports, just like the istio-pilot charm's code is inherently tied to the versions *it* supports. Specifying the version in the library is equivalent to the charm specifying the versions in it's `metadata.yaml` and doing it here keeps it closest to, and coupled with, the code that actually has that compatibility requirement. If we offload it to the consuming charm, then they can put whatever versions they want and the library will break.

The schema URL default, OTOH, is just convenience so that the consuming charm authors don't have to copypasta that when they're already copypastaing the entire library in via `charmcraft`.  Also, the intention is that pretty much everyone should add the `local_sdi` bit to their `charmcraft.yaml` so that it doesn't have external runtime dependencies, which is why it allows for the `metadata.yaml` to override it.

mmanciop:
New issue: setup stuck on `List of ingress versions not found for apps: prometheus`:

```
michele@boombox:~/git/prometheus-operator$ juju show-unit prometheus/0
prometheus/0:
  opened-ports: []
  charm: local:focal/prometheus-k8s-0
  leader: true
  relation-info:
  - endpoint: grafana-source
    related-endpoint: grafana-source
    application-data: {}
    related-units:
      grafana/0:
        in-scope: true
        data:
          egress-subnets: 10.152.183.251/32
          ingress-address: 10.152.183.251
          private-address: 10.152.183.251
  - endpoint: grafana-dashboard
    related-endpoint: grafana-dashboard
    application-data: {}
    related-units:
      grafana/0:
        in-scope: true
        data:
          egress-subnets: 10.152.183.251/32
          ingress-address: 10.152.183.251
          private-address: 10.152.183.251
  - endpoint: ingress
    related-endpoint: ingress
    application-data:
      _supported_versions: |-
        - v1
        - v3
        - v2
    related-units:
      istio-pilot/0:
        in-scope: true
        data:
          egress-subnets: 10.152.183.151/32
          ingress-address: 10.152.183.151
          private-address: 10.152.183.151
  provider-id: prometheus-0
  address: 10.1.151.113
michele@boombox:~/git/prometheus-operator$ juju show-unit istio-pilot/0
istio-pilot/0:
  opened-ports: []
  charm: local:focal/istio-pilot-0
  leader: true
  relation-info:
  - endpoint: istio-pilot
    related-endpoint: istio-pilot
    application-data:
      _supported_versions: '- v1'
    related-units:
      istio-gateway/0:
        in-scope: true
        data:
          egress-subnets: 10.152.183.39/32
          ingress-address: 10.152.183.39
          private-address: 10.152.183.39
  - endpoint: ingress
    related-endpoint: ingress
    application-data:
      service-hostname: prometheus
      service-name: prometheus
      service-port: "9090"
    related-units:
      prometheus/0:
        in-scope: true
        data:
          egress-subnets: 10.152.183.179/32
          ingress-address: 10.152.183.179
          private-address: 10.152.183.179
  provider-id: istio-pilot-0
  address: 10.1.151.105
```

johnsca:
@mmanciop That seems to have sent the data for the nginx `ingress` relation protocol rather than the istio one:

```yaml
  - endpoint: ingress
    related-endpoint: ingress
    application-data:
      service-hostname: prometheus
      service-name: prometheus
      service-port: "9090"
```

mmanciop:
> I added a `MockIngressProvider` to the library for use with unit tests, as well as a unit test for the library itself which uses it. Hopefully that makes testing charms which use the library easier.

That is super useful, thank you! Although, it seems not to work for unit tests in which `self.harness.set_leader(False)` is invoked.

johnsca:
Oof, thanks for the catch.  I added coverage in the library test for the non-leader case (which also covers the `MockIngressProvider`) and fixed that issue.

mmanciop:
For the record: I have yet to see the routing work yet :-(

mmanciop:
> For the record: I have yet to see the routing work yet :-(
> 
> Everything seems roughly in place in terms of resources. I see istio serving the 404s (there is a telling header in the response), but no luck routing.

Turns out it was not an issue with the proxy, but with Prometheus generating the wrong redirect URLs :-)

mmanciop:
When removing the relation between `prometheus` and `istio-pilot`, I still get this though:

```
unit-prometheus-0: 12:33:05 ERROR unit.prometheus/0.juju-log ingress:2: Uncaught exception while in charm code:
  File "/var/lib/juju/agents/unit-prometheus-0/charm/venv/ops/model.py", line 1521, in _run
    result = run(args, **kwargs)
  File "/usr/lib/python3.8/subprocess.py", line 512, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '('/var/lib/juju/tools/unit-prometheus-0/relation-get', '-r', '2', '-', '', '--app', '--format=json')' returned non-zero exit status 1.

During handling of the above exception, another exception occurred:

  File "./src/charm.py", line 378, in <module>
    main(PrometheusCharm)
  File "/var/lib/juju/agents/unit-prometheus-0/charm/venv/ops/main.py", line 414, in main
    charm = charm_class(framework)
  File "./src/charm.py", line 71, in __init__
    endpoint_address=self._remote_write_address,
  File "./src/charm.py", line 374, in _remote_write_address
    return self._external_hostname if self.model.get_relation("ingress") else None
  File "./src/charm.py", line 367, in _external_hostname
    if unit_external_url := self.ingress.unit_url:
  File "/var/lib/juju/agents/unit-prometheus-0/charm/lib/charms/istio_pilot/v0/ingress.py", line 286, in unit_url
    return (self.unit_urls or {}).get(self.charm.unit.name)
  File "/var/lib/juju/agents/unit-prometheus-0/charm/lib/charms/istio_pilot/v0/ingress.py", line 277, in unit_urls
    return self._data.get("unit_urls")
  File "/usr/lib/python3.8/functools.py", line 967, in __get__
    val = self.func(instance)
  File "/var/lib/juju/agents/unit-prometheus-0/charm/lib/charms/istio_pilot/v0/ingress.py", line 242, in _data
    if not self._sdi:
  File "/usr/lib/python3.8/functools.py", line 967, in __get__
    val = self.func(instance)
  File "/var/lib/juju/agents/unit-prometheus-0/charm/lib/charms/istio_pilot/v0/ingress.py", line 159, in _sdi
    return SerializedDataInterface(
  File "/var/lib/juju/agents/unit-prometheus-0/charm/venv/serialized_data_interface/__init__.py", line 110, in __init__
    others = {
  File "/var/lib/juju/agents/unit-prometheus-0/charm/venv/serialized_data_interface/__init__.py", line 111, in <dictcomp>
    app.name: bag.get("_supported_versions")
  File "/usr/lib/python3.8/_collections_abc.py", line 660, in get
    return self[key]
  File "/var/lib/juju/agents/unit-prometheus-0/charm/venv/ops/model.py", line 430, in __getitem__
    return self._data[key]
  File "/var/lib/juju/agents/unit-prometheus-0/charm/venv/ops/model.py", line 414, in _data
    data = self._lazy_data = self._load()
  File "/var/lib/juju/agents/unit-prometheus-0/charm/venv/ops/model.py", line 779, in _load
    return self._backend.relation_get(self.relation.id, self._entity.name, self._is_app)
  File "/var/lib/juju/agents/unit-prometheus-0/charm/venv/ops/model.py", line 1588, in relation_get
    return self._run(*args, return_output=True, use_json=True)
  File "/var/lib/juju/agents/unit-prometheus-0/charm/venv/ops/model.py", line 1523, in _run
    raise ModelError(e.stderr)
ops.model.ModelError: b'ERROR "" is not a valid unit or application\n'
```

johnsca:
Per discussion at the sprint last week, due to the difficulties in getting Istio to work with this (admittedly not exactly intended) usage pattern, as well as some concerns with how Istio handles some things like intermittent backend outages, and the complications that the per-unit routing adds to the otherwise straightforward base use-case for the `ingress` relation, we are pivoting to Traefik for per-unit ingress and splitting `ingress-unit` into its own interface.  Keeping this PR as a draft for now until [the Traefik charm](https://github.com/canonical/traefik-k8s-operator/pull/1) is ready to have the `ingress-unit` library migrated to it.

DomFleischmann:
Given that the observability team has moved to using traefik for this I believe we can safely close this PR

PR #47, opened 2021-12-17
.github/workflows/publish.yaml: fix matrix syntax

DnPlas:
* charms/*/metadata.yaml: specifies image latest tag 
* .github/workflows/publish.yaml: fix matrix syntax


DomFleischmann (approved):
:cake:

PR #49, opened 2022-01-06
fix ingress-auth add tests, and other changes

ca-scribner:
Fixes the ingress-auth relation, which used CRDs that were removed from this version of Istio. Unit tests pass, but not yet fully integration tested with dex/oidc-gatekeeper

Incorporates the following changes from johnsca's [outstanding PR](https://github.com/canonical/istio-operators/pull/43):
- Splits out default_gateway handling from ingress handling.  Adds/refactors tests for this
- ~~adds destination rule creation and~~ refactors remove hook (edit: destination rules removed)

Fixes bug when using multiple ingress relation where the template rendering merged the yaml for the different virtualservices together

Fixes possible bug where, previously, when handling relation changes we deleted all VirtualService, EnvoyFilters, etc. that were created by this juju app, and then recreated the objects we wanted. This had the unintended consequence of removing any objects that istiod (which is also deployed by this application) created. This fix adds an extra label to identify istio entities created by this charm's workload from those created by other sources within the application

(from knkski) Fixes case where charm fails if ingress auth routes list is empty.  Previously,
`manifests` would never be empty, due to `rbac_configs`. With that removal, `manifests` might be empty. In which case, we skip creating it.


charms/istio-pilot/src/charm.py
```diff
@@ -179,25 +209,57 @@ def handle_ingress_auth(self, event):
...
+
+        If the gateway isn't available or doesn't have a load balancer address yet,
+        returns None.
+        """
+        svcs = yaml.safe_load(
+            self._kubectl(
+                "get",
+                "svc",
+                "-l",
+                "istio=ingressgateway",
```
DomFleischmann, line 251: Is this a typo? Should this be `istio-ingressgateway`?

ca-scribner: I believe this is correct.  I think it is trying to select objects created by the istio-gateway charm in the [manifest](https://github.com/canonical/istio-operators/blob/ca-scribner/fix-ingress-auth-main/charms/istio-gateway/src/manifest.yaml), which have an `istio={{kind}}gateway` label

charms/istio-gateway/pyproject.toml
```diff
@@ -0,0 +1,4 @@
+[tool.black]
```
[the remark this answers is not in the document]
ca-scribner: well that's interesting, I didn't realise black looks upward in the tree to find this.  I assumed you needed this file in `/charms/some_charm` to make running lint in that `some_charm` dir work, but you don't.  Top level works just fine.  Removed.

charms/istio-pilot/src/charm.py
```diff
@@ -130,54 +132,91 @@ def remove(self, event):
...
-        gateways = ''.join(t.render(name=g) for g in gateways)
-
-        manifests = [virtual_services, gateways]
-        manifests = '\n'.join([m for m in manifests if m])
-
-        for resource in [self.virtual_service_resource, self.gateway_resource]:
-            self._delete_existing_resource_objects(resource, namespace=self.model.name)
+        self._delete_existing_resource_objects(
+            self.virtual_service_resource, namespace=self.model.name
+        )
```
DomFleischmann, line 216: Am I understanding correctly that we are deleting all the virtual services and recreating them each time we call `handle_ingress`? Is this necessary?

DomFleischmann (commented):
This looks promising, just some things that I don't understand completely.

charms/istio-pilot/src/charm.py
```diff
@@ -130,54 +132,91 @@ def remove(self, event):
...
+            self._get_gateway_address
+        except (ApiError, TypeError) as e:
+            if e == ApiError:
+                self.log.exception("ApiError: Could not get istio-ingressgateway, retrying")
+            elif e == TypeError:
+                self.log.exception("TypeError: No ip address found, retrying")
+            event.defer()
+            return
+        else:
+            self.unit.status = ActiveStatus()
```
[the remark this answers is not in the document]
DnPlas: Oh, that was left from the rebase. I'll rebase it.

charms/istio-pilot/src/charm.py
```diff
@@ -266,6 +313,18 @@ def _delete_manifest(
                 ignore_unauthorized=ignore_unauthorized,
             )
 
+    @property
+    def _get_gateway_address(self):
```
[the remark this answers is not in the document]
DnPlas: Yeh, I inherited the name from the old PR, I can change it to `_gateway_address`

charms/istio-pilot/src/charm.py
DnPlas: Yes, we are deleting and creating VirtualServices every time. Regarding the need for it, I think @ca-scribner can elaborate more. I'm not sure if it is because we didn't have an `apply` option in lightkube. 
Also, there's an important concern raised [here](https://github.com/canonical/istio-operators/issues/50) regarding the implications of doing it this way, if we want to get rid of it, that could be a good reason.

DomFleischmann: Yes that was my concern too, specially considering that some charms might be very noisy with this relation.

charms/istio-pilot/src/charm.py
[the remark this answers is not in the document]
ca-scribner: Status setting in general in this charm uses some of the practices we decided were problematic (`__init__` sets status to Active, etc.).  It is out of scope for fixing the ingress-auth, but if we're going to use this charm probably we should update the status setting (but as a separate PR, not as part of this PR)

charms/istio-pilot/src/charm.py
ca-scribner: Yeah I think this was more challenging before using lightkube and so Ken used this as a easy first implementation.

Lightkube's `apply` makes `add` and `update` actions really easy, but we also need to cover `delete`.  For an `ingress` relation that is dropped, our relation data bag just wont have that relation, so doing a simple `apply` of the metadata would update new/existing `VirtualServices` but would just leave `VirtualService`s for removed relations in place (since an `apply` never deletes anything).  But again with lightkube, this is pretty easy to cover - one solution would be:
* lightkube.get all VirtualServices **created by this charm** (we don't want to delete virtualservices from other stuff like seldon or knative)
* compare existing `VirtualService`s to desired, deleting any that exist now but are not backed by an existing relation
* apply the metadata for new/existing

ca-scribner: (Identifying which `vs`s are from this charm vs other sources was a problem in the past too and is addressed in this charm by adding the (poorly named - please rename if you think of something better) label `app.istio-pilot.io/is-workload-entity=true`)

charms/istio-pilot/src/charm.py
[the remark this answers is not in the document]
DnPlas: I agree on setting the status, but in a more intelligent way. I can work on that in a follow up PR.

charms/istio-pilot/src/charm.py
DnPlas: I'd like to explore the option of getting the `virtualservices` created by this charm, comparing desired against existing ones without a backing relation (delete those) and apply new/existing. That is addressing a different issue, though, I'd like to have a separate PR just for that.

ca-scribner:
Good news: I think these changes generally work.  I can access the kubeflow bundle via an ingress with auth, etc., and it looks good!  I made a few changes to testing and CI but nothing to the main code.  If someone can look those over that would be great!

Bad news: but that took a lot of effort because there's existing problems with the charm that are not addressed here (eg: problems that existed before this update):
* status handling is deficient (mentioned in previous comments).  For example, deploy the `istio-gateway` charm without setting the `kind` config variable and the charm will be `Active` but looking at `show-status-log` we see it is actually `Blocked`
* some of the relation logic and event triggering needs to be rethought.  For example try this: 
  1. deploy istio-pilot
  1. relate some ingresses to istio-pilot
  1. relate istio-pilot to istio-ingressgateway (required for virtualservices to be created)
  1. <-- no virtual services will be created (the handle_ingress hook doesn't get triggered)
* (mentioned above already) `VirtualService`s are deleted any time we handle_ingress, leading to (brief) disruptions
* when istio-pilot cannot find an ingress gateway (which it looks for by a hard-coded service name), it doesn't notify anyone of the problem.  This came up for me when my ingressgateway wasn't firing up (because I forgot the `kind=ingress`, and thus I had no virtualservices, but both pilot and gateway were active with no obvious bad log messages

I think it makes sense to merge the current PR and handle all this in separate issues, but that everything noted here and in the other comments should be completed before we use this in the kubeflow bundle.  Thoughts @DnPlas @DomFleischmann ?

ca-scribner:
CI still fails in github during integration tests.  This might be a false positive where the ingressgateway controller reports as blocked while it waits for istio-pilot?  Needs further inspection...

PR #53, opened 2022-01-20
Replace kubectl with lightkube

stonepreston:
This PR removes the use of kubectl from the gateway and pilot charms and replaces it with lightkube.


charms/istio-gateway/src/charm.py
```diff
@@ -79,12 +84,25 @@ def remove(self, event):
...
-        subprocess.run(
-            ["./kubectl", "delete", "-f-"],
-            input=rendered.encode('utf-8'),
-            # Can't remove stuff yet: https://bugs.launchpad.net/juju/+bug/1941655
-            # check=True
-        )
+        try:
+            for obj in codecs.load_all_yaml(rendered):
+                self.lightkube_client.delete(
+                    obj.__class__, obj.metadata.name, namespace=obj.metadata.namespace
```
addyess, line 90: do you like `obj.__class__` or `type(obj)` better here?

charms/istio-gateway/tests/conftest.py
```diff
@@ -0,0 +1,75 @@
+import pytest
```
addyess, line 1: this looks like the beginning of a unit test framework?  if so, could you push all the unit tests into ```.../tests/unit/conftest.py``` so there's room for integration tests (whenever they should come?). If there is already integration tests elsewhere -- feel free to ignore

charms/istio-gateway/tests/conftest.py
```diff
@@ -0,0 +1,75 @@
...
+class Helpers:
+    @staticmethod
+    def get_unique_calls(call_args_list):
+        uniques = []
+        for call in call_args_list:
+            if call in uniques:
+                continue
+            else:
+                uniques.append(call)
+        return uniques
```
addyess, line 16: is `call_args_lists` hashable (maybe its a tuple of arguments?)    If so you should be able to 
```suggestion
   def get_unique_calls(call_args_list):
       return set(call_args_list)
```

which kinda makes you think this method isn't necessary

charms/istio-gateway/tests/test_charm.py
```diff
@@ -21,95 +15,70 @@ def test_no_kind(harness):
...
-        actual_input = list(yaml.safe_load_all(call.kwargs['input']))
-        assert call.args == expected_args
-        assert expected_input == actual_input
-
-    assert harness.charm.model.unit.status == ActiveStatus('')
+def test_install_apply(configured_harness, kind, mocked_client, helpers):
+    actual_objects = []
+    expected_objects = list(yaml.safe_load_all(open(f'tests/{kind}-example.yaml')))
+
+    # The install method is invoked multiple times, and the apply method is called for every object
```
addyess, line 30: i'm not opposed to testing with mocks this way.  
more often i see the patterns [here](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called)

```python
mocked_client.return_value.apply.assert_called_with(...)
mocked_client.return_value.apply.assert_not_called(...)
mocked_client.return_value.apply.assert_has_calls(...)
```

etc.  Any reason you went about iterating against the call list?

charms/istio-gateway/tests/test_charm.py
```diff
@@ -21,95 +15,70 @@ def test_no_kind(harness):
...
-    harness.begin_with_initial_hooks()
-
-    container = harness.model.unit.get_container('noop')
-    harness.charm.on['noop'].pebble_ready.emit(container)
+    helpers.begin_noop(harness)
 
     assert harness.charm.model.unit.status == BlockedStatus('Waiting for istio-pilot relation')
 
 
-def test_kind_ingress(harness, mocker):
```
addyess, line 36: are these deleted tests no longer valid?

charms/istio-pilot/src/charm.py
```diff
@@ -181,23 +214,57 @@ def handle_ingress_auth(self, event):
...
-        )
+        for resource in [self.envoy_filter_resource, self.rbac_config_resource]:
+            self._delete_existing_resource_objects(resource, namespace=self.model.name)
+        self._apply_manifest(manifests, namespace=self.model.name)
+
+    def _delete_object(
+        self, obj, namespace=None, ignore_not_found=False, ignore_unauthorized=False
+    ):
+        try:
+            self.lightkube_client.delete(obj.__class__, obj.metadata.name, namespace=namespace)
```
addyess, line 226: here's another shot for `type(obj)` over `obj.__class__`

charms/istio-pilot/src/charm.py
```diff
@@ -181,23 +214,57 @@ def handle_ingress_auth(self, event):
...
+        except ApiError as err:
+            self.log.error("ApiError encountered while attempting to delete resource.")
+            if err.status.message is not None:
+                if "not found" in err.status.message and ignore_not_found:
+                    self.log.error(f"Ignoring not found error:\n{err.status.message}")
+                elif "(Unauthorized)" in err.status.message and ignore_unauthorized:
+                    # Ignore error from https://bugs.launchpad.net/juju/+bug/1941655
+                    self.log.error(f"Ignoring unauthorized error:\n{err.status.message}")
+                else:
+                    self.log.error(err.status.message)
```
addyess, line 236: would you consider using so the stacktrace bubbles out in the logs too?  although, you are re-raising so i guess it's entirely possible someone out of this scope might do the same thing.   ¯\_(ツ)_/¯
```python
self.log.exception(err.status.message)
```

charms/istio-pilot/src/charm.py
```diff
@@ -181,23 +214,57 @@ def handle_ingress_auth(self, event):
...
+                raise
+
+    def _delete_existing_resource_objects(
+        self, resource, namespace=None, ignore_not_found=False, ignore_unauthorized=False
+    ):
+        for obj in self.lightkube_client.list(
+            resource, labels={"app.juju.is/created-by": f"{self.app.name}"}
+        ):
+            self._delete_object(
+                obj,
```
addyess, line 248: is this `obj` or is `type(obj)`.  If it really is meant to be `klass` -- i suggest a rename here in this loop

but when I [`read-the-docs`](https://lightkube.readthedocs.io/en/latest/client/#list-or-watch-objects-in-all-namespaces) it seems `client.list(...)` does return the entire object.  In other circumstances you've used `type(obj)` during deletions.  Can you verify this is doing the right thing?

charms/istio-pilot/src/charm.py
```diff
@@ -181,23 +214,57 @@ def handle_ingress_auth(self, event):
...
+    def _apply_manifest(self, manifest, namespace=None):
+        for obj in codecs.load_all_yaml(manifest):
+            self.lightkube_client.apply(obj, namespace=namespace)
+
+    def _delete_manifest(
+        self, manifest, namespace=None, ignore_not_found=False, ignore_unauthorized=False
+    ):
+        for obj in codecs.load_all_yaml(manifest):
+            self._delete_object(
+                obj,
```
addyess, line 263: same here.  before you used `type(obj)` not `obj`.

charms/istio-pilot/tests/conftest.py
```diff
@@ -0,0 +1,86 @@
+import pytest
```
addyess, line 1: Please see my above comments regarding unittest vs integration tests.

charms/istio-pilot/tests/conftest.py
```diff
@@ -0,0 +1,86 @@
...
+from charm import Operator
+from ops.testing import Harness
+
+
+class Helpers:
+    @staticmethod
+    def get_unique_calls(call_args_list):
```
addyess, line 8: please see my above comments about hashable items and sets and how to `mock_obj.assert_*` methods

addyess (changes requested):
This is a huge undertaking and it looked necessary.  Nice work.  I have a few questions and would probably changed things around places where you use dunder ```__class__``` and ```__name___``` to get at some very private traits of an object.  Check through some of the places where you use `lightkube_client.delete` for use of `obj` rather than `type(obj)` and re-hightlight me when you've got your next review up

charms/istio-gateway/tests/test_charm.py
stonepreston: the deleted tests were combined into one paramterized test with the kind feature. The egress/ingress tests were the same test, with the difference being one used the ingress example file and the other used the egress file.

charms/istio-pilot/tests/conftest.py
```diff
@@ -0,0 +1,86 @@
...
+def subprocess(mocker):
+    subprocess = mocker.patch("charm.subprocess")
+    for method_name in ("run", "call", "check_call", "check_output"):
+        method = getattr(subprocess, method_name)
+        method.return_value.returncode = 0
+        method.return_value.stdout = b""
+        method.return_value.stderr = b""
+        method.return_value.output = b""
+        mocker.patch(f"subprocess.{method_name}", method)
+    yield subprocess
```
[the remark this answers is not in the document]
stonepreston: Subprocess is still used for istioctl calls, these could probably be whittled down a bit though since I dont think run is used at all anymore

charms/istio-pilot/src/charm.py
stonepreston: so the lightkube delete method takes a class as its first argument, the resource kind. As its second argument it takes the name of the object that is to be deleted. This felt weird to me so I added the _delete_object helper method that just takes in an object, which seemed less confusing than having to pass in a resource type and object name explicitly.

addyess: ooo!  i see this now.  thanks for clearing that up.

charms/istio-gateway/tests/conftest.py
stonepreston: TypeError: unhashable type: '_Call'
Not hashable :/

charms/istio-gateway/src/charm.py
stonepreston: I've replaced uses of __class__ with type() calls

charms/istio-gateway/tests/conftest.py
stonepreston: The tests for both charms have been refactored into their own unit/ directory within tests/

charms/istio-gateway/tests/conftest.py
stonepreston: I refactored the tests such that the non-unique calls is no longer an issue and was able to get rid of this helper function entirely.

charms/istio-pilot/src/charm.py
stonepreston: Ive incorporated self.log.exception into the initial log statement

charms/istio-pilot/tests/conftest.py
stonepreston: Moved the tests into a /unit directory

charms/istio-pilot/tests/conftest.py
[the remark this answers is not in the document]
stonepreston: I removed the run and call methods from the mocks since they are no longer used. check_call and check_output are both used for istioctl

charms/istio-gateway/tests/test_charm.py
addyess: yeah `assert_has_calls` is the closest -- but t just ensures that a subset of the calls were made in a specified order

addyess (approved):
LGTM @stonepreston

charms/istio-gateway/tests/conftest.py
johnsca: The integration tests live at the top (bundle) level because they depend on (and cover) both charms.

johnsca (approved):
LGTM

PR #55, opened 2022-01-21
Add custom gateway option

### LINKED ISSUE #54: Allow to configure our own default gateway
Hi there, I am using juju to deploy the entire Kubeflow clusters.
I want to use my own gateway configuration to handle the ingress.
However, whenever I changed the gateway's configuation, juju will help me to revert back to the original one.
Is there any way for me to change the default gateway's configuration?

RyanSiu1995:
Close #54.
It is useful in most of the cases.
For example, we can use a custom gateway to configure the TLS and HTTP redirect.


RyanSiu1995:
Hi @DomFleischmann, do you think you can review this PR?

DomFleischmann:
Hi @RyanSiu1995 thank you for the contribution this completely ran under my radar. I'll add @knkski and @johnsca as reviewers.

charms/istio-pilot/config.yaml
```diff
@@ -3,3 +3,7 @@ options:
     default: istio-gateway
     description: Comma-separated list of gateways to create by default
     type: string
+  namespaced-custom-gateways:
+    default: ""
+    description: Comma-separated list of namespaced custom gateways to use
+    type: string
```
johnsca, line 9: I'm not sure that it's useful to include "namespaced" in the option name and it makes it a bit unwieldy.  Suggestion:
```suggestion
  custom-gateways:
    default: ""
    description: >-
      Comma-separated list of custom or manually configured Gateway resources to use.
      Names can include the namespace, in the form "namespace/gateway", or defaults to
      the model in which the charm is deployed.
    type: string
```

(Note: The default for the namespace is [handled automatically by Istio](https://istio.io/latest/docs/reference/config/networking/virtual-service/#VirtualService-gateways), and the explicit namespace in the current template is actually superfluous.)

charms/istio-pilot/src/charm.py
```diff
@@ -117,8 +117,14 @@ def get_kwargs(version, route):
 
             return kwargs
 
+        custom_gateway = (
+            []
+            if self.model.config['namespaced-custom-gateways'] == ''
+            else self.model.config['namespaced-custom-gateways'].split(',')
+        )
```
johnsca, line 124: A bit more concise:
```suggestion
        custom_gateways = filter(None, self.config["custom-gateways"].split(","))
```

charms/istio-pilot/src/charm.py
```diff
@@ -117,8 +117,14 @@ def get_kwargs(version, route):
...
+
         virtual_services = ''.join(
-            t.render(**get_kwargs(ingress.versions[app.name], route))
+            t.render(**get_kwargs(ingress.versions[app.name], route), custom_gateway=custom_gateway)
```
johnsca, line 127:
```suggestion
            t.render(**get_kwargs(ingress.versions[app.name], route), custom_gateways=custom_gateways)
```

charms/istio-pilot/src/virtual_service.yaml.j2
```diff
@@ -6,6 +6,9 @@ metadata:
 spec:
   gateways:
   - '{{ namespace }}/{{ gateway }}'
+  {% for i in custom_gateway %}
```
johnsca, line 9:
```suggestion
  {% for i in custom_gateways %}
```

charms/istio-pilot/tests/test_charm.py
```diff
@@ -240,3 +240,116 @@ def test_with_ingress_auth_relation(harness, mocker):
...
+    rel_id = harness.add_relation("ingress", "app")
+
+    harness.add_relation_unit(rel_id, "app/0")
+    data = {"service": "service-name", "port": 6666, "prefix": "/"}
+    harness.update_relation_data(
+        rel_id,
+        "app",
+        {"_supported_versions": "- v1", "data": yaml.dump(data)},
+    )
+    harness.update_config({"namespaced-custom-gateways": ",".join(gateways)})
```
johnsca, line 280:
```suggestion
    harness.update_config({"custom-gateways": ",".join(gateways)})
```

johnsca (commented):
Mostly LGTM but a few small suggestions which are not necessarily blockers if Ken disagrees.

charms/istio-pilot/src/virtual_service.yaml.j2
knkski: Could we do `gateways = [gateway] + custom_gateways` in the charm code, then include them in the template as `{{ gateways | tojson | safe }}`?

RyanSiu1995: Yes, sure!

charms/istio-pilot/config.yaml
RyanSiu1995: Updated.

charms/istio-pilot/src/charm.py
RyanSiu1995: Updated. Thank you for this tip!

charms/istio-pilot/src/charm.py
RyanSiu1995: Updated

charms/istio-pilot/src/virtual_service.yaml.j2
RyanSiu1995: After trying to the aggregated way, I still think the original is much simpler.
The aggregated one needs to handle the logic of `{{ namespace }}/{{ gateway }}`
It makes the code inside the charm code really complicated because the namespace is extracted inside a loop.
Therefore, I prefer to keep the current approach so that it looks simpler (but of course looks dumb too)

charms/istio-pilot/tests/test_charm.py
RyanSiu1995: Updated.

RyanSiu1995:
Thank you all! I have updated the codes according to the comment.
Please have a review again to see if it makes sense to you guys 🙇🏻

RyanSiu1995:
Hi folks, how should I fix the integration test here?
I cannot see any error popping up here.

PR #56, opened 2022-01-21
Fix status handling and make the charms buildable

DomFleischmann:
- Centralize all the logic in to one method
- Pin dependencies to pull correct libraries
- Add files to charmcraft.yaml

### DIFF
--- a/charms/istio-ingressgateway/src/charm.py
+++ b/charms/istio-ingressgateway/src/charm.py
@@ -16,21 +16,6 @@ from oci_image import OCIImageResource, OCIImageResourceError
 class Operator(CharmBase):
     def __init__(self, *args):
         super().__init__(*args)
-        if not self.unit.is_leader():
-            # We can't do anything useful when not the leader, so do nothing.
-            self.model.unit.status = WaitingStatus("Waiting for leadership")
-            return
-
-        try:
-            self.interfaces = get_interfaces(self)
-        except NoVersionsListed as err:
-            self.model.unit.status = WaitingStatus(str(err))
-            return
-        except NoCompatibleVersions as err:
-            self.model.unit.status = BlockedStatus(str(err))
-            return
-        else:
-            self.model.unit.status = ActiveStatus()
 
         self.log = logging.getLogger(__name__)
         self.image = OCIImageResource(self, "oci-image")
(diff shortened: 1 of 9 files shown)


charms/istio-ingressgateway/src/charm.py
```diff
@@ -232,5 +243,16 @@ def check_ca_root_cert(self, namespace):
         return config_map
 
 
+class CheckFailed(Exception):
+    """ Raise this exception if one of the checks in main fails. """
+
+    def __init__(self, msg, status_type=None):
+        super().__init__()
+
+        self.msg = msg
```
ca-scribner, line 252: Handle messages that are not strings (for example, passing exceptions directly).  This came up in some of our other charms.

```suggestion
        self.msg = str(msg)
```

charms/istio-pilot/src/charm.py
```diff
@@ -364,6 +358,37 @@ def set_pod_spec(self, event):
...
+        return image_details
+
+
+class CheckFailed(Exception):
+    """ Raise this exception if one of the checks in main fails. """
+
+    def __init__(self, msg, status_type=None):
+        super().__init__()
+
+        self.msg = msg
```
ca-scribner, line 388: Same as above
```suggestion
        self.msg = str(msg)
```

charms/istio-pilot/src/charm.py
```diff
@@ -17,59 +17,53 @@ class Operator(CharmBase):
...
+        if interfaces["istio-pilot"]:
+            interfaces["istio-pilot"].send_data(
                 {
                     "service-name": f'{self.model.app.name}.{self.model.name}.svc',
                     "service-port": str(self.model.config['xds-ca-tls-port']),
                 }
             )
+            self.log.info("istio-pilot relation is available info has been sent")
+        else:
+            self.log.info("istio-pilot relation not available not able to send service info")
```
ca-scribner, line 44:
```suggestion
            self.log.info("istio-pilot relation is available.  Service data written to relation")
        else:
            self.log.info("istio-pilot relation not available.  No service data written to relation")
```

ca-scribner: Just cleaning up language

charms/istio-pilot/test/unit/test_charm.py
```diff
@@ -76,3 +76,35 @@ def test_main_service_mesh(harness):
...
+#     relation.data[relation.app]["_supported_versions"] = "- v1"
+#     # relation.data[relation.app]["data"] = yaml.dump(data)
+#     # harness.update_relation_data(
+#     #     rel_id,
+#     #     "app",
+#     #     {"_supported_versions": "- v1", "data": yaml.dump(data)},
+#     # )
+#     harness.charm.on.istio_pilot_relation_joined.emit(relation)
+#
+#     assert relation.data[harness.charm.app] == None
```
ca-scribner, line 110: Do we want to keep this in here?  I don't have an opinion, just making sure it isn't here by mistake

PR #59, opened 2022-02-17
bundle.yaml: renames istio-ingressgateway -> istio-gateway in relation

DnPlas:
The was renamed `istio-ingressgateway`  -> `istio-gateway` recently, this change should also be reflected in the relation.


DnPlas:
Closing this as it does not apply.

Issue #42, opened 2021-11-30
istio-pilot fails to deploy


VariableDeclared:
Hi 

When deploying the istio bundle, the following error is thrown:
```
creating or updating custom resource definitions: ensuring custom resource definit ion "meshpolicies.authentication.istio.io" with version "v1beta1": the server could not find the requested resource
```

Kubernetes version: 1.22.4-3+adc4115d990346 (Microk8s)
Deployed using repo instructions

Thanks

Peter

DomFleischmann:
The version available on charmstore is `istio 1.5` which does not support kuberenetes 1.22 due to the API removals. The new istio oprator which is currently merged in Github uses `istio 1.11` and supports 1.22. It is not yet entirely published on charmhub but you can build it locally and try it out

VariableDeclared:
thanks

PR #61, opened 2022-02-23
Pin markupsafe==2.0.1 for Jinja bug workaround

jardon:
Pin Jinja dependency markupsafe to v2.0.1

This issue is caused by a change in Jinja's dependencies.  This is a workaround that forces an older version of markupsafe for the time being.

https://github.com/pallets/jinja/issues/1585

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -27,7 +27,7 @@ jobs:
 
     - uses: balchua/microk8s-actions@v0.2.2
       with:
-        channel: 'latest/stable'
+        channel: '1.21/stable'
         addons: '["dns", "storage", "rbac", "metallb:10.64.140.43-10.64.140.49"]'
 
     - name: Install dependencies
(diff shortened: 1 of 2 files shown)


phvalguima:
Hi, the integration test is failing because ```istio-pilot``` is stuck waiting for a CRD that never comes up. The unit itself shows an error on:
```
istio-pilot           istio/pilot:1.5.0  error        1  istio-pilot    local             0  kubernetes           creating or updating custom resource definitions: ensuring custom resource definition "meshpolicies.authentication.istio.io" with version "v1beta1": the server could not find the requested resource
```

Looking for the bugs on istio-operators, I can see #42, which seems to be a similar issue.

I'd say if you run the step ```Run balchua/microk8s-actions@v0.2.2``` but pinning to ```channel=1.21/stable``` instead of latest/stable.

ca-scribner (approved):

PR #64, opened 2022-03-09
Improves bundle testing and removes oci-image resource

DnPlas:
* charms/*/metadata.yaml: replaces oci-image with noop
* tests/test_bundle.py: replaces build_bundle with explicit & individual methods
    juju-bundle is not as transparent as using individual commands, and since this is a test, we can afford the extra steps, so we can be more explicit on what we are executing and what part is causing issues in case of a failure.

Depends on #63


DomFleischmann (commented):
This looks good but I just merged the other PR which made the CI work, can you resolve the conflicts and run the tests again?

PR #65, opened 2022-03-09
istio-pilot/src/charm.py: implements a reconcile resources method

### LINKED ISSUE #50: Does deleting and recreating virtualservice objects on relation change interrupt workloads?
On a ingress relation change, we currently delete all virtualservices and then recreate the desired virtualservices.  This results in a (short) period where no virtualservices exist (while short, it still persists for a second or two and can be caught by `watch -n 1 kubectl get vs` during a relation change event).  Will this cause problems for the users of the ingress/virtualservice?

One possible problem that comes to mind is for kubeflow, where users of notebook servers may be interrupted and need to reconnect.  This has not yet been tested.

DnPlas:
In previous versions of this file we are removing all resources of
a kind on different events and then applying them again. This could
raise issues if any related charm is using such resource at the moment
of removal.
Using a reconcile helper method allows us to keep alive the existing
resources and only remove those that are not backed by a relation.
In this commit this method is implemented and also used for the
handle_ingress event handler as a fix for #50.

Fixes #50

### DIFF
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -2,6 +2,7 @@
 
 import logging
 import subprocess
+from typing import Any
 
 import yaml
 from jinja2 import Environment, FileSystemLoader
@@ -161,7 +162,7 @@ class Operator(CharmBase):
 
     def handle_ingress(self, event):
         try:
-            self._get_gateway_address
+            self._gateway_address
         except (ApiError, TypeError) as e:
             if e == ApiError:
                 self.log.exception("ApiError: Could not get istio-ingressgateway, retrying")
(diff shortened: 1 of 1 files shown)


ca-scribner:
I like this a lot.  My only real comment is related to what you said in the daily today about making this a reusable pattern.  I think this is close to being a (actually, two!) nice upstream contribution to lightkube.  Not that we have to push this upstream immediately, but describing things in upstreamed-terms as I think it is clearer:

* I think `_reconcile_desired_resources` could be refactored a little and made into a completely generic helper that could become `lightkube.client.reconcile(res, namespace, labels, desired_resources)`.  Adding `labels` into the params would make it really flexible, and then `_apply_manifest` could be lightkube-ified in line with how they'd do it upstream (maybe if upstream `desired_resources` would be a list of objects rather than the raw yaml?).  I think I'm missing a piece or two without actually coding it, but I'm **pretty sure** it would work nicely
* lightkube has a `client.deletecollection(resource, namespace)` which deletes groups of objects.  We could extend that to accept a `labels` dict and it could be a near drop-in replacement for `_delete_existing_resources_object`

As a first step, you could move these lightkube-like functions either to a lightkube fork or just to a separate file as helpers (just to make really sure things are separated cleanly).  What do you think?

DnPlas:
@ca-scribner I agree, this could even be part of upstream lightkube. I'll take a two step approach with this one:
1) Extract all the lightkube related bits, `_reconcile_desired_resources` included, from the charm code into a helper module to ensure these changes get merged soon.
2) Push the "reconcile" logic to upstream lightkube, wait for it to get merged (if accepted), and then migrate the helper module.

I'm closing this PR and opening a new one with the proposed changes.

PR #68, opened 2022-03-14
Refactors charm.py extracting all lightkube logic to a helper module

DnPlas:
* istio-pilot/src: extracts all lightkube logic to a helper module
* istio-pilot/tests/unit: skipping all tests as they need to be refactored

> Note: unit tests will be changed in a follow up PR

### DIFF
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -5,14 +5,14 @@ import subprocess
 
 import yaml
 from jinja2 import Environment, FileSystemLoader
+from lightkube import Client
+from lightkube.core.exceptions import ApiError
+from lightkube.resources.core_v1 import Service
 from ops.charm import CharmBase, RelationBrokenEvent
 from ops.main import main
 from ops.model import ActiveStatus, BlockedStatus, WaitingStatus
 from serialized_data_interface import NoCompatibleVersions, NoVersionsListed, get_interfaces
-from lightkube import Client, codecs
-from lightkube.core.exceptions import ApiError
-from lightkube.generic_resource import create_namespaced_resource
-from lightkube.resources.core_v1 import Service
+from resources_handler import ResourceHandler, GenericNamespacedResource
 
 
 class Operator(CharmBase):
@@ -37,43 +37,11 @@ class Operator(CharmBase):
 
         self.log = logging.getLogger(__name__)
 
-        # Every lightkube API call will use the model name as the namespace by default
-        self.lightkube_client = Client(namespace=self.model.name, field_manager="lightkube")
-        # Create namespaced resource classes for lightkube client
-        # This is necessary for lightkube to interact with custom resources
-        self.envoy_filter_resource = create_namespaced_resource(
-            group="networking.istio.io",
-            version="v1alpha3",
-            kind="EnvoyFilter",
-            plural="envoyfilters",
-            verbs=None,
-        )
-
-        self.virtual_service_resource = create_namespaced_resource(
-            group="networking.istio.io",
-            version="v1alpha3",
-            kind="VirtualService",
-            plural="virtualservices",
-            verbs=None,
-        )
-
-        self.gateway_resource = create_namespaced_resource(
-            group="networking.istio.io",
-            version="v1beta1",
-            kind="Gateway",
-            plural="gateways",
-            verbs=None,
-        )
-
-        self.rbac_config_resource = create_namespaced_resource(
-            group="rbac.istio.io",
-            version="v1alpha1",
-            kind="RbacConfig",
-            plural="rbacconfigs",
-            verbs=None,
-        )
-
         self.env = Environment(loader=FileSystemLoader('src'))
+        self._custom_resource_classes = {}
+        self._resource_handler = ResourceHandler(self.app.name, self.model.name)
+
+        self.lightkube_client = Client(namespace=self.model.name, field_manager="lightkube")
 
         self.framework.observe(self.on.install, self.install)
         self.framework.observe(self.on.remove, self.remove)
@@ -81,7 +49,6 @@ class Operator(CharmBase):
         self.framework.observe(self.on.config_changed, self.handle_default_gateway)
 
         self.framework.observe(self.on["istio-pilot"].relation_changed, self.send_info)
-
         self.framework.observe(self.on['ingress'].relation_changed, self.handle_ingress)
         self.framework.observe(self.on['ingress'].relation_broken, self.handle_ingress)
         self.framework.observe(self.on['ingress'].relation_departed, self.handle_ingress)
@@ -120,15 +87,11 @@ class Operator(CharmBase):
             ]
         )
 
-        for resource in [
-            self.virtual_service_resource,
-            self.gateway_resource,
-            self.envoy_filter_resource,
-        ]:
-            self._delete_existing_resource_objects(
+        for resource in self._custom_resource_classes.values():
+            self._resource_handler.delete_existing_resource_objects(
                 resource, namespace=self.model.name, ignore_unauthorized=True
             )
-        self._delete_manifest(
+        self._resource_handler.delete_manifest(
             manifests, namespace=self.model.name, ignore_not_found=True, ignore_unauthorized=True
         )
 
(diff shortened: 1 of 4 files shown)


charms/istio-pilot/tests/unit/test_charm.py
```diff
@@ -6,6 +6,7 @@
 from lightkube.core.exceptions import ApiError
 
 
+@pytest.mark.skip("Skipping as unit tests need to be refactored")
```
DomFleischmann, line 9: I would prefer if this PR includes the refactoring of the unit tests.

DomFleischmann (commented):
This looks promising but I have some comments.

charms/istio-pilot/src/resources_handler.py
```diff
@@ -0,0 +1,103 @@
...
+
+import yaml
+from jinja2 import Environment, FileSystemLoader
+from lightkube import Client, codecs
+from lightkube.core.exceptions import ApiError
+from lightkube.generic_resource import create_namespaced_resource, GenericNamespacedResource
+
+
+class ResourceHandler:
+    def __init__(self, config):
```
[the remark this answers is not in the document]
DnPlas: I left it because originally I was receiving more than just `app_name` and `model_name`,things such as the context for rendering manifests, paths, etc. But probably I can simplify this, as you said, just receiving those two. I'll change it.

charms/istio-pilot/src/charm.py
```diff
@@ -256,62 +217,48 @@ def handle_ingress_auth(self, event):
...
+        # improve how we do this.
+        context = {
+            'namespace': 'namespace',
+            'app_name': 'name',
+            'name': 'generic_resource',
+            'request_headers': 'request_headers',
+            'response_headers': 'response_headers',
+            'port': 'port',
+            'service': 'service',
+        }
```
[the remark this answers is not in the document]
DnPlas: You are right, I'll move it to the handler code

charms/istio-pilot/src/resources_handler.py
```diff
@@ -0,0 +1,118 @@
...
+    ):
+        for obj in codecs.load_all_yaml(manifest):
+            self.delete_object(
+                obj,
+                namespace=namespace,
+                ignore_not_found=ignore_not_found,
+                ignore_unauthorized=ignore_unauthorized,
+            )
+
+    def create_ns_resource(self, filename: str, context: dict = None) -> GenericNamespacedResource:
```
ca-scribner, line 88: I think after a refactor to provide default values for `context`, the `context` in the function args is unused.  Is the intent to use the below defaults always, or just when the function is called without some of the fields?

ca-scribner: Unrelated to above, when reading through code that invokes `create_ns_resource` I interpreted it as creating the **resource** (in k8s), not the Lightkube resource object that represents the resource.  I'm not sure how to improve this, but if you can think of something to help that would be great.  I realise this name probably derives from Lightkube calling their own functions `create_namespaced_resource`, so maybe it makes sense to leave it, but still something to consider.

Also, if you're really making a wrapper around the lightkube function, I'd consider either calling it the same name (`create_namespaced_resource`) or something meaningfully different (`create_lightkube_generic_resource_class`).  Shortening the `namespaced` to `ns` but leaving it the same felt like an awkward middle ground.  But that's just one person's opinion :)

charms/istio-pilot/src/charm.py
```diff
@@ -256,62 +217,35 @@ def handle_ingress_auth(self, event):
...
-            resource, labels={"app.juju.is/created-by": f"{self.app.name}"}.update(labels)
-        ):
-            self._delete_object(
-                obj,
-                namespace=namespace,
-                ignore_not_found=ignore_not_found,
-                ignore_unauthorized=ignore_unauthorized,
-            )
+    @property
+    def _custom_resources(self) -> dict:
```
ca-scribner, line 226: I don't think it really matters because this likely isn't called really frequently, but in general I think it would be good practice to avoid (re-)creating all resources whenever the user asks for any generic resource class.  It looks like Lightkube will pass you the existing resource if you've already made the same thing once before so that's good, but we still create extras on most passes (if I want virtual service, I still create envoy filter and gateway) and if I need another virtual service later I still read all the yamls and try to create things a second time

A possible approaches could be:
* using a getter rather than a normal property, creating whatever people ask as they're needed remembering the things you've already created and just returning them, something like:
```python
def __init__(self):
  ...
  self._custom_resource_classes = {}
def get_custom_resource_class(self, name: str) -> Resource:
  if name in self._custom_resource_classes:
    return self._custom_resource_classes[name]
  else:
    # Create the resource class
    ...
    self._custom_resource_classes[name] = new_resource_class
    return self._custom_resource_classes[name]
```

charms/istio-pilot/src/charm.py
```diff
@@ -141,14 +102,14 @@ def handle_default_gateway(self, event):
         t = self.env.get_template('gateway.yaml.j2')
         gateway = self.model.config['default-gateway']
         manifest = t.render(name=gateway, app_name=self.app.name)
-        self._delete_existing_resource_objects(
-            resource=self.gateway_resource,
+        self._resource_handler.delete_existing_resource_objects(
+            resource=self._custom_resources['gateway'],
             labels={
-                "app.juju.is/created-by": f"{self.app.name}",
                 "app.{self.app.name}.io/is-workload-entity": "true",
```
ca-scribner, line 108: This is a mistake from before this PR, but this string should be an f-string
```suggestion
                f"app.{self.app.name}.io/is-workload-entity": "true",
```

Is the `juju/created-by` label being removed because of redundancy?

charms/istio-pilot/src/resources_handler.py
```diff
@@ -0,0 +1,118 @@
...
+                raise
+
+    def delete_existing_resource_objects(
+        self,
+        resource,
+        namespace=None,
+        ignore_not_found=False,
+        ignore_unauthorized=False,
+        labels={},
+    ):
```
ca-scribner, line 60: (from #69)
```suggestion
        labels=None,
    ):
        if labels is None:
            labels = {}
        
```

ca-scribner (changes requested):
In general things are looking really good.  Added some suggestions along the way

The only other thing I noticed was that `resources_handler.py` could use some unit tests.  It is nicely broken up into bite-sized functions that can easily be tested in isolation - we should add those tests here as well.  There are probably other little bugs somewhere in the details of those, but they're hard to find without isolating each piece.

charms/istio-pilot/src/resources_handler.py
DnPlas: Agree, will import some of those changes here.

charms/istio-pilot/src/charm.py
```diff
@@ -256,62 +217,35 @@ def handle_ingress_auth(self, event):
...
+            'envoy_filter': envoy_filter_resource,
+            'virtual_service': virtual_service_resource,
+            'gateway': gateway_resource,
+        }
+        return custom_resources
+
+    @property
+    def _resource_handler(self):
+        """Returns an instanciated ResourceHandler."""
+        return ResourceHandler(self.app.name, self.model.name)
```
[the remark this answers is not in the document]
DnPlas: I see. If we really think about it, we could even do this:

```python
    def __init__(...):
      ...
      self._resource_handler = ResourceHandler(self.app.name, self.model.name)
```

charms/istio-pilot/src/charm.py
DnPlas: +1 on the typo, the juju label is a default in the methods that depend on labels inside `resource_handler.py`, that's because we always care about the resources that are created by juju.

charms/istio-pilot/src/resources_handler.py
DnPlas: Regarding your first comment, yes `context` is an unused arg. I recently changed this bit, removed the arg from the docstrings, but forgot the args list in the method definition. Will change that, good catch.

Regarding your second comment, yeh, I also felt awkward naming this method. I didn't want it to be named exactly the same as the Lightkube method to avoid confusion. I'd rather have something meaningfully different, in fact `get_generic_resource_class` or a shorter `generic_resource_class` could be more accurate as the method actually returns the resource class. What do you think?

charms/istio-pilot/src/charm.py
DnPlas: I agree.

charms/istio-pilot/src/resources_handler.py
DnPlas: Changed it from `create_ns_resource` to `generate_generic_resource_class`, let me know if that is better.

ca-scribner (approved):

PR #69, opened 2022-03-15
Implements reconcile resources method

### LINKED ISSUE #50: Does deleting and recreating virtualservice objects on relation change interrupt workloads?

DnPlas:
istio-pilot/src/resource_handler.py: implements a reconcile resources method
    
In previous versions of this file we are removing all resources of
a kind and then applying them again w/o caring about their origin.
This could raise issues if any related charm is using such resource at the moment
of removal.
Using a reconcile helper method allows us to keep alive the existing
resources and only remove those that are not backed by a relation.
In this commit this method is implemented and also used for the
handle_ingress event handler as a fix for #50.
 Fixes #50

This PR also adds unit tests for the reconcile method, but those depend on #74

### DIFF
--- a/charms/istio-pilot/src/resources_handler.py
+++ b/charms/istio-pilot/src/resources_handler.py
@@ -4,12 +4,15 @@
 
 """Resources handling library using Lightkube."""
 import logging
+from typing import Tuple, TextIO, Union, Iterable
 
 import yaml
 from jinja2 import Environment, FileSystemLoader
+import lightkube  # noqa F401  # Needed for patching in test_resources_handler.py
 from lightkube import Client, codecs
 from lightkube.core.exceptions import ApiError
 from lightkube.generic_resource import create_namespaced_resource, GenericNamespacedResource
+from lightkube.core.resource import Resource
 
 
 class ResourceHandler:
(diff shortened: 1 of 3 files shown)


charms/istio-pilot/src/resources_handler.py
```diff
@@ -0,0 +1,152 @@
...
+            resource,
+            labels={
+                "app.juju.is/created-by": f"{self.app_name}",
+                f"app.{self.app_name}.io/is-workload-entity": "true",
+            },
+            namespace=namespace,
+        )
+        if desired_resources:
+            desired_resources = codecs.load_all_yaml(desired_resources)
+            diff_obj = set(existing_resources) - set(desired_resources)
```
ca-scribner, line 141: These objects are not hashable.  Might need to do something more nuanced to get a diff?  I'm seeing these in the debug-log:
```
  File "./src/charm.py", line 266, in <module>
    main(Operator)
  File "/var/lib/juju/agents/unit-istio-pilot-0/charm/venv/ops/main.py", line 424, in main
    framework.reemit()
  File "/var/lib/juju/agents/unit-istio-pilot-0/charm/venv/ops/framework.py", line 746, in reemit
    self._reemit()
  File "/var/lib/juju/agents/unit-istio-pilot-0/charm/venv/ops/framework.py", line 783, in _reemit
    custom_handler(event)
  File "./src/charm.py", line 177, in handle_ingress
    self._resource_handler.reconcile_desired_resources(
  File "/var/lib/juju/agents/unit-istio-pilot-0/charm/src/resources_handler.py", line 141, in reconcile_desired_resources
    diff_obj = set(existing_resources) - set(desired_resources)
TypeError: unhashable type: 'VirtualService'
```

charms/istio-pilot/src/resources_handler.py
```diff
@@ -0,0 +1,198 @@
...
+            else:
+                raise
+
+    def delete_existing_resource_objects(
+        self,
+        resource,
+        namespace=None,
+        ignore_not_found=False,
+        ignore_unauthorized=False,
+        labels=None,
```
ca-scribner, line 62: This was `labels={}`, but I changed it to `labels=None` with a later `if labels is Non:, labels = {}` which I think is what you mean to use.  Having `labels=AnyNewMutableObject` has the (usually unexpected) behaviour of making `labels` global across all calls of `delete_existing_resource_objects`, which I don't think you intended.

charms/istio-pilot/src/resources_handler.py
```diff
@@ -0,0 +1,198 @@
...
+    ):
+        if labels is None:
+            labels = {}
+
+        for obj in self.lightkube_client.list(
+            resource,
+            labels={"app.juju.is/created-by": f"{self.app_name}"}.update(labels),
```
ca-scribner, line 69: I think this might be a remnant from when this was not a generic function.  Maybe remove this so it is truly generic?  

If you'd like a way of doing convenience labels across all calls of a ResourceHandler, you could always add an optional arg to `__init__` that way it is controllable.

charms/istio-pilot/src/resources_handler.py
```diff
@@ -0,0 +1,198 @@
...
+            version=manifest["apiVersion"].split("/")[1],
+            kind=manifest["kind"],
+            plural=f"{manifest['kind']}s".lower(),
+            verbs=None,
+        )
+
+        return ns_resource
+
+    def reconcile_desired_resources(
+        self, resource: Any, desired_resources: Union[str, TextIO, None], namespace: str = None
```
ca-scribner, line 127: I'm not sure if the `namespace` arg works as expected.  The docstring mentions it is optional, but is it?  Does `lightkube_client.list(..., namespace=None)` work?  I'm expecting no, but maybe that just gets non-namespaced resources?  I'm also not sure if `.apply(..., namespace=None)` works or not (maybe it takes namespace from the Resource objects themselves, or it is for non-namespaced objects?).  In any case, if you're not sure either, maybe make namespace a required arg

ca-scribner:
In resources_handler.py (and maybe elsewhere?) the terms `object` and `resource` seem interchangeable (eg `.delete_object()`, `.reconcile_desired_resources()`, ...).  If that's the case, maybe settle on just one to help with clarity.  I think lightkube usually uses `resource`

charms/istio-pilot/src/resources_handler.py
DnPlas: I think this is resolved with your latest changes.

charms/istio-pilot/src/resources_handler.py
DnPlas: Left is as non-optional.

DnPlas:
> In resources_handler.py (and maybe elsewhere?) the terms `object` and `resource` seem interchangeable (eg `.delete_object()`, `.reconcile_desired_resources()`, ...). If that's the case, maybe settle on just one to help with clarity. I think lightkube usually uses `resource`

Will do, I will be cleaning up the code in a follow up PR.

charms/istio-pilot/src/resources_handler.py
DnPlas: Solved in #68

PR #75, opened 2022-03-30
fix: istio-pilot error handling around gateway address

DnPlas:
Previously handle_ingress checked for errors by equality `(caught_error == SomeErrorType)`, which asserts that the caught_error is actually the class not an instance of the class. This is fixed here by changing to `isinstance(caught_error, SomeErrorType)`

### DIFF
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -98,8 +98,8 @@ class Operator(CharmBase):
     def handle_default_gateway(self, event):
         """Handles creating gateways from charm config
 
-        Side effect: self.handle_ingress() is also invoked by this handler as ingress objects
-        depend on the default_gateway
+        Side effect: self.handle_ingress() is also invoked by this handler as ingress
+        objects depend on the default_gateway
         """
         t = self.env.get_template('gateway.yaml.j2')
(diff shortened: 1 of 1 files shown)


charms/istio-pilot/src/charm.py
```diff
@@ -124,12 +124,26 @@ def send_info(self, event):
 
     def handle_ingress(self, event):
         try:
-            self._get_gateway_address
+            if self._get_gateway_address:
+                self.unit.status = ActiveStatus()
```
DomFleischmann, line 128: Can we comfortably set an active status here if this statement is true?

ca-scribner: I think that makes sense.  And possibly at the top of this charm there could also be a `MaintenanceStatus`, but I haven't thought that through much so I'll go with your gut on whether it is needed.  

fwiw, this charm still has an `ActiveStatus` in the `__init__` so all these nuances will get trampled whenever an unattached hook fires, but that's a future problem for us :)

DnPlas: I suggest we think about statuses better and discuss them in a follow up PR.

DomFleischmann (approved):

PR #78, opened 2022-04-05
istio-pilot/src: refactors delete helper methods

DnPlas:
Previously, we were not considering the case of deleting resources
created by other entities than lightkube, causing the remove event
handler to raise LoadResourceError exceptions (through helpers).
This commit handles that case and also refactors a couple other
helper methods to suit these needs.

### DIFF
--- a/charms/istio-pilot/src/resources_handler.py
+++ b/charms/istio-pilot/src/resources_handler.py
@@ -10,7 +10,7 @@ import yaml
 from jinja2 import Environment, FileSystemLoader
 import lightkube  # noqa F401  # Needed for patching in test_resources_handler.py
 from lightkube import Client, codecs
-from lightkube.core.exceptions import ApiError
+from lightkube.core.exceptions import LoadResourceError, httpx
 from lightkube.generic_resource import create_namespaced_resource, GenericNamespacedResource
 from lightkube.core.resource import Resource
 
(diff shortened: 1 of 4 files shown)


charms/istio-pilot/src/resources_handler.py
```diff
@@ -35,30 +35,34 @@ def __init__(self, app_name, model_name):
...
+        # FIXME: find out why this is the case, probably something missing
+        # in upstream lightkube
+        except httpx.HTTPStatusError as err:
+            self.log.warning(
+                "HTTPStatusError encountered while attempting to delete resource.\n"
+                "If ignore_not_found is set to True, this WARNING can be dismissed."
+            )
+            if err.response.status_code == 404 and ignore_not_found:
+                self.log.warning(f"Ignoring not found error for: {obj.kind}/{obj.metadata.name}")
+                return
```
ca-scribner, line 57: Not specifically something you're changing, but should this be a `return` here?  It feels more like it should be one of:
* `continue`: If this is a warning and might happen due to a bad object in a yaml, but it is ignorable and we should keep going
* `raise`: If this a real error and things should stop right away

charms/istio-pilot/src/resources_handler.py
```diff
@@ -72,31 +76,55 @@ def delete_existing_resources(
...
 
         Args:
-            - filename: name of the manifest file used to create the resource
+            - manifest: Kubernetes resource object used to create the resource
+            - from_filename (optional): when True, it creates generic resources
+                  from a manifest file instead of an object
         """
 
+        # FIXME: this method receives two mutually exclusive args
+        # we should change that
```
ca-scribner, line 126: Is this correct?  Below `manifest` gets used `if from_filename`

charms/istio-pilot/src/resources_handler.py
```diff
@@ -72,31 +76,55 @@ def delete_existing_resources(
...
+                )
+                return
+            try:
+                obj_to_delete = lightkube.codecs.from_dict(obj)
+            except LoadResourceError:
+                if include_created_by_external:
+                    self.generate_generic_resource_class(manifest=obj)
+                    obj_to_delete = lightkube.codecs.from_dict(obj)
+                else:
+                    raise
```
ca-scribner, line 106: Can you explain what this does?  I'm not sure what the `LoadResourceError` would be indicating.

DnPlas:
Closing this PR as these changes are tracked in #84 and #85

PR #82, opened 2022-04-06
Pseudo-code that demonstrates explicitly defining custom resources charms

ca-scribner:


DnPlas:
Thanks @ca-scribner . Closing this PR as we agreed on the implementation for `resource_handler.py`.

PR #83, opened 2022-04-22
new relation to pass gateway info

agathanatasha:
New gateway relation that passes gateway name and gateway namespace to the related application.
Gateway relation data is updated when istio-pilot's default-gateway config is updated

Related knative-istio-controller PR: https://github.com/canonical/knative-operators/pull/7


charms/istio-pilot/tests/unit/test_charm.py
```diff
@@ -287,6 +287,23 @@ def test_with_ingress_auth_relation(harness, subprocess, helpers, mocked_client,
...
 
+def test_correct_data_in_gateway_relation(harness, subprocess):
+    harness.set_leader(True)
+
+    rel_id = harness.add_relation("gateway", "app")
+    harness.add_relation_unit(rel_id, "app/0")
+    harness.set_model_name("test-model")
+    harness.begin_with_initial_hooks()
+    print("DEBUGGING test harness model app", harness.model.app)
+    print("DEBUGGING test harness model name", harness.model.name)
```
ca-scribner, line 298: Debug code that can be removed?

```suggestion
```

charms/istio-pilot/src/charm.py
```diff
@@ -261,6 +266,12 @@ def handle_ingress_auth(self, event):
         )
         self._apply_manifest(auth_filters, namespace=self.model.name)
 
+    def handle_gateway_relation(self, event):
+        relations = [rel for rel in self.model.relations["gateway"] if rel.app]
+        for relation in relations:
+            relation.data[self.app]["gateway-name"] = self.model.config["default-gateway"]
+            relation.data[self.app]["gateway-namespace"] = self.model.name
```
ca-scribner, line 273: What data is provided by this relation if the charm does not have a gateway (eg: if it is blocked because it cannot create one)?  I think it will either never populate the relation with data (which is kinda ok, but means the other side of the relation needs to check for empty data) or it will respond with the `default-gateway` name even if istio isn't actually working properly.  I don't think either is quite what you're looking for

ca-scribner: I wonder if `handle_gateway_relation` should have a helper `get_gateway` and use that rather than `self.model.config` for the gateway name.  That could then return None if no gateway is up.  

Also, should we return a single gateway name and namespace here, or should we return a list of gateways?  If we think we might handle more than one in future, making this a list now might save us pain later

ca-scribner: @DomFleischmann I don't actually know how to interact with relation data directly so you'll need to check that.  in particular, I wasn't sure how the `if rel.app` part worked

charms/istio-pilot/tests/unit/test_charm.py
agathanatasha: removed, thanks for catching that

charms/istio-pilot/src/charm.py
agathanatasha: > @DomFleischmann I don't actually know how to interact with relation data directly so you'll need to check that. in particular, I wasn't sure how the `if rel.app` part worked

It doesn't, will update

DomFleischmann (changes requested):
I would still want most of the logic of this and [this PR](https://github.com/canonical/knative-operators/pull/7) as a charmcraft library.

charms/istio-pilot/metadata.yaml
```diff
@@ -17,5 +17,9 @@ provides:
     interface: ingress-auth
     schema: https://raw.githubusercontent.com/canonical/operator-schemas/master/ingress-auth.yaml
     versions: [v1]
+  gateway:
+    interface: gateway
```
DomFleischmann, line 21: This is a very generic name, I don't think we can confidently create a gateway interface which could be listed in this [repository](https://github.com/canonical/charm-relation-interfaces). So lets be more specific with the name of the interface, e.g. `istio-gateway-name`

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info_receiver.py
```diff
@@ -0,0 +1,55 @@
+import logging
```
DomFleischmann, line 1: With charmcraft libraries , each file is a new library. Given that we want to implement one library for this interface, both of them should be in the same file.

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info_receiver.py
```diff
@@ -0,0 +1,55 @@
...
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 1
+
+
+DEFAULT_RELATION_INTERFACE = "gateway"
+
+
+class GatewayConsumer(Object):
```
DomFleischmann, line 16: Although I agree that Consumer is a more intuitive name we should probably stick with Requirer so that we use the same terminology as in the juju metadata.

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info_receiver.py
```diff
@@ -0,0 +1,55 @@
...
+        self.charm = charm
+        self.log = logging.getLogger(__name__)
+        self.relation_name = relation_name
+
+    def get_relation_data(self):
+        if not self.model.unit.is_leader():
+            return
+        gateway = self.model.relations[self.relation_name]
+        if len(gateway) == 0:
+            raise Exception("Missing gateway relation with istio-pilot")
```
DomFleischmann, line 28: I would like to see a custom exception instead of just an `Exception` this makes it basically impossible to filter during a `try ... except` statement.

DomFleischmann (changes requested):

charms/istio-pilot/src/charm.py
```diff
@@ -124,7 +126,7 @@ def handle_default_gateway(self, event):
         )
         self._resource_handler.apply_manifest(manifest)
 
-        # Update the ingress resources as they rely on the default_gateway
+        # Update the ingress objects as they rely on the default_gateway
```
DnPlas, line 129: We had an agreement to call them `resources` instead of `objects`. See [this](https://github.com/canonical/istio-operators/commit/312b6d027317f2dab2cf6947150ef628f2cac270) commit.

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_name.py
```diff
@@ -0,0 +1,121 @@
...
+        self.charm = charm
+        self.resource_handler = resource_handler
+        self.framework.observe(
+            charm.on[DEFAULT_RELATION_NAME].relation_changed, self._on_gateway_relation_changed
+        )
+        self.framework.observe(charm.on.config_changed, self._on_gateway_config_changed)
+        self.framework.observe(charm.on.update_status, self._on_gateway_config_changed)
+
+    def _validate_gateway_exists(self):
+        response = self.lightkube_client.get(
```
DnPlas, line 97: I'd add a try/catch block here, this `lightkube_client.get()` could end up raising errors that will prevent this response variable to be defined. 
Also, thinking about this method a bit more, `lightkube_client.get()` returns an object that belongs to a class, in this case the class of the resource you are looking for, `Gateway`. You could do something very redundant, but also very explicit (since this is a library, it should be okay) like the following:

```python
from lightkube.generic_resource import create_namespaced_resource
...
def __init__(...):
  self.gateway_class = create_namespaced_resources("define the Gateway resource here")

def _validate_gateway_exists(self):
  try:
    gateway_resource = client.get(self.gateway_class, self.model.config['default-gateway'], namespace=self.model.name)
  except ApiError:
   some_error_handling
  if isinstance(gateway_resource, gateway_class):
    return True
  return False
```
This helps you to catch any Exceptions the lightkube client might throw and also avoids the dependency on resource_handler.
If you don't want to do a redundant check, you can leave the `get` and return a True if no exceptions are raised, False otherwise.

DnPlas (commented):
Thanks @agathanatasha ! This is looking good, just a couple of comments.

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_name.py
```diff
@@ -0,0 +1,175 @@
...
+                        "gateway_name": self.model.config["default-gateway"]
+                        if self._validate_gateway_exists()
+                        else "",
+                        "gateway_namespace": self.model.name,
+                    }
+                )
+
+    def _on_gateway_config_changed(self, event):
+        if len(self.model.relations[DEFAULT_RELATION_NAME]) > 0:
+            self._on_gateway_relation_changed(event)
```
ca-scribner, line 175: I don't know if we need this.  I think any event triggering `_on_gateway_config_changed` could also trigger `_on_gateway_relation_changed`, or at least directly trigger it without additional logic here (since `_on_gateway_relation_changed` can handle cases with zero relations)

ca-scribner:
@DomFleischmann I think my initial instincts might have been different from what you suggested above (I see some comments above asking for more logic in the library vs the charm), so figured I'd ask about your thinking.  

When reading through this it makes me think I'm still not sure about the purpose of a library.  The `GatewayProvider` is implemented with pretty intimate details about the charm itself (uses lightkube to `.get` a gateway named `charm.config['default-gateway']`, etc).  This feels like charm code, not a shared library.  What is your thought on this?  I'd have thought the library would encapsulate things like how we encode/decode the relation data, and the charm would know how to figure out if it had a gateway up right now.

Separately, the more logic we put into the library, the more dependencies we pass onto consumers of the library.  For example, atm consumers need `lightkube` just because the `GatewayProvider` (which they don't use) needs `lightkube`.  If we're putting this logic in the library, perhaps we should at least separate it to different files.  Alternatively, we could also put imports inside our classes, but that's generally bad practice (although in this case, it should work)

DomFleischmann:
> When reading through this it makes me think I'm still not sure about the purpose of a library. The `GatewayProvider` is implemented with pretty intimate details about the charm itself (uses lightkube to `.get` a gateway named `charm.config['default-gateway']`, etc). This feels like charm code, not a shared library. What is your thought on this? I'd have thought the library would encapsulate things like how we encode/decode the relation data, and the charm would know how to figure out if it had a gateway up right now.

After giving this some further thought, I think the main issue with the current implementation, as you correctly pointed out, is that we are checking if the gateway is up in the wrong place. This should be handled through the charm code, as it is the one that is creating that resource. 

How I would approach this is by creating a custom event in the gateway library, e.g. `GatewayAvailableEvent`. I would then observe that specific event in the library instead of `update-status` and remove any kind of checks for the gateway in the library. In the charm code itself, it already checks if the gateway is available, therefore once it is I would emit the `GatewayAvailableEvent` too which the library will react and update the relation data.

This should solve our dependency issue and make this library more reusable.

@ca-scribner @agathanatasha @DnPlas thoughts?

charms/istio-pilot/src/istio_gateway_name_provider.py
```diff
@@ -0,0 +1,25 @@
...
+logger = logging.getLogger(__name__)
+
+DEFAULT_RELATION_NAME = "gateway"
+
+
+class GatewayProvider(Object):
+    def __init__(self, charm, relation_name=DEFAULT_RELATION_NAME):
+        super().__init__(charm, relation_name)
+
+    def on_gateway_relation_changed(self, charm, gateway_name, gateway_namespace):
```
ca-scribner, line 16: I think this name might be from when this actually handled the relation changed event.  Now it feels more like a `send_data` or something like that.

charms/istio-pilot/src/istio_gateway_name_provider.py
```diff
@@ -0,0 +1,25 @@
...
+        if self.model.unit.is_leader():
```
ca-scribner, line 17: I don't know if I prefer the is_leader check here or in the charm itself.  If you haven't thought about it since moving this out of a library, give it some though and go with whichever you prefer.

Regardless of where it is, we should do some logging if not leader.  As is, I could see it being really hard to debug if the charm does `gateway_object.on_gateway_relation_changed(...)` and sees no exceptions, no logging, and no change in the relation data.  I'm not sure if this should raise an exception that is caught by the charm, or just log that no data was sent.  But something should happen

charms/istio-pilot/src/istio_gateway_name_provider.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def send_gateway_relation_data(self, charm, gateway_name, gateway_namespace):
+        relations = self.model.relations["gateway"]
+        for relation in relations:
+            relation.data[charm].update(
+                {
+                    "gateway_name": gateway_name,
+                    "gateway_namespace": gateway_namespace,
+                }
+            )
```
DomFleischmann, line 21: Why was the leadership check removed entirely? As I understand, this part of the code will fail when a non-leader executes it.

charms/istio-pilot/src/charm.py
```diff
@@ -127,6 +132,22 @@ def handle_default_gateway(self, event):
...
+            resource_type=self._resource_handler.get_custom_resource_class_from_filename(
+                "gateway.yaml.j2"
+            ),
+            resource_name=self.model.config['default-gateway'],
+            resource_namespace=self.model.name,
+        )
+        if is_gateway_created:
+            self.gateway.send_gateway_relation_data(
+                self.app, self.model.config['default-gateway'], self.model.name
+            )
```
DomFleischmann, line 149: Maybe a log message for when the data is not sent would be helpful.

DomFleischmann (changes requested):
This seems almost ready to be merged, just 2 details

charms/istio-pilot/src/istio_gateway_name_provider.py
agathanatasha: There is a leadership check in operator init

DomFleischmann (approved):

PR #84, opened 2022-04-22
Changes logic on istio-pilot to use lightkube implicit resource creation

DnPlas:
This PR introduces the following changes to be able to use Lightkube's experimental feature of implicitly creating generic resources.
* istio-pilot/src: changes logic to use implicit resource creation
* istio-pilot: adds requirements and config to pull LK from branch


charms/istio-pilot/src/resources_handler.py
```diff
@@ -90,16 +90,16 @@ def delete_manifest(
                 ignore_unauthorized=ignore_unauthorized,
             )
 
-    def generate_generic_resource_class(self, filename: str) -> GenericNamespacedResource:
+    def get_custom_resource_class_from_filename(self, filename: str) -> GenericNamespacedResource:
```
ca-scribner, line 93: This can return a Global resource as well

charms/istio-pilot/charmcraft.yaml
```diff
@@ -9,6 +9,7 @@ bases:
 parts:
   charm:
     charm-python-packages: [setuptools, pip]
+    build-packages: [git]
```
ca-scribner, line 12: I don't know if I've had to use this before - what does this add?

DnPlas: We need it to be able to fetch [this](https://github.com/ca-scribner/lightkube/tree/add-implicit-model-generation) experimental LK version

charms/istio-pilot/src/resources_handler.py
DnPlas: Yeh, is `Any` good enough?

ca-scribner: could also have a `Union[]`, but this is good

ca-scribner (approved):
lgtm

PR #88, opened 2022-05-03
Add SSL support

phvalguima:
Add ssl-* configs to pass cert and key. That should be
used to create a secret and added to the gateway yaml.

Added --diff to lint to show more details on tox.ini.

### DIFF
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -114,12 +114,14 @@ class Operator(CharmBase):
         resources depend on the default_gateway
         """
         # Clean-up resources
-        self._delete_existing_resource_objects(
-            resource=self.gateway_resource,
+        self._resource_handler.delete_existing_resources(
+            resource=self._resource_handler.get_custom_resource_class_from_filename(
+                filename='gateway.yaml.j2'
+            ),
             labels={
-                "app.juju.is/created-by": f"{self.app.name}",
-                "app.{self.app.name}.io/is-workload-entity": "true",
+                f"app.{self.app.name}.io/is-workload-entity": "true",
             },
+            namespace=self.model.name,
         )
         t = self.env.get_template('gateway.yaml.j2')
         gateway = self.model.config['default-gateway']
@@ -130,14 +132,6 @@ class Operator(CharmBase):
         )
         manifest = None
         if secret_name:
-            manifest = t.render(
-                name=gateway,
-                secret_name=secret_name,
-                ssl_crt=self.model.config["ssl-crt"],
-                ssl_key=self.model.config["ssl-key"],
-                model_name=self.model.name,
-                app_name=self.app.name,
-            )
             secret = self.env.get_template('gateway-secret.yaml.j2')
             manifest_secret = secret.render(
                 secret_name=secret_name,
@@ -146,11 +140,13 @@ class Operator(CharmBase):
                 model_name=self.model.name,
                 app_name=self.app.name,
             )
-            self._apply_manifest(manifest_secret)
-            return
+            self._resource_handler.apply_manifest(manifest_secret)
 
         manifest = t.render(
             name=gateway,
+            secret_name=secret_name,
+            ssl_crt=self.model.config["ssl-crt"] or None,
+            ssl_key=self.model.config["ssl-key"] or None,
             model_name=self.model.name,
             app_name=self.app.name,
         )
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -113,11 +113,18 @@ class Operator(CharmBase):
         Side effect: self.handle_ingress() is also invoked by this handler as ingress
         resources depend on the default_gateway
         """
-
+        # Clean-up resources
+        self._delete_existing_resource_objects(
+            resource=self.gateway_resource,
+            labels={
+                "app.juju.is/created-by": f"{self.app.name}",
+                "app.{self.app.name}.io/is-workload-entity": "true",
+            },
+        )
         t = self.env.get_template('gateway.yaml.j2')
         gateway = self.model.config['default-gateway']
         secret_name = (
-            "istio-gw-secret"
+            f"{self.app.name}-gateway-secret"
             if self.model.config["ssl-crt"] and self.model.config["ssl-key"]
             else None
         )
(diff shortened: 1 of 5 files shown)


charms/istio-gateway/metadata.yaml
```diff
@@ -5,6 +5,7 @@ description: |
   https://istio.io/latest/docs/tasks/traffic-management/ingress/
   https://istio.io/latest/docs/tasks/traffic-management/egress/
 min-juju-version: "2.9.0"
+series: [kubernetes]
```
ca-scribner, line 8: I thought this would change the charm to a podspec charm, which would have prevented lightkube from working at all (due to RBAC).  Have you tested with rbac enabled?

The CI I think would fail here, but there's conflicts with main and it wont run until they're resolved

charms/istio-pilot/metadata.yaml
```diff
@@ -4,6 +4,7 @@ summary: |
   Istio Service Mesh.
 description: |
   https://istio.io/latest/docs/reference/commands/pilot-discovery/
+series: [kubernetes]
```
ca-scribner, line 7: same as above

charms/istio-pilot/src/gateway.yaml.j2
```diff
@@ -11,7 +12,27 @@ spec:
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ secret_name }}
+  namespace: {{ model_name }}
+type: kubernetes.io/tls
+data:
+  tls.crt: {{ ssl_crt }}
+  tls.key: {{ ssl_key }}
+{% endif %}
```
ca-scribner, line 38: This seems a bit unexpected that `gateway.yaml.j2` has the side effect of sometimes including a secret.  Can we put the secret in a separate `gateway-secret.yaml.j2` or similar, and selectively apply that file?

charms/istio-pilot/src/charm.py
```diff
@@ -140,7 +140,27 @@ def handle_default_gateway(self, event):
         """
         t = self.env.get_template('gateway.yaml.j2')
         gateway = self.model.config['default-gateway']
-        manifest = t.render(name=gateway, app_name=self.app.name)
+        secret_name = (
+            "istio-gw-secret"
```
ca-scribner, line 144: Can we name this dynamically based on the charm name, like `"{charm-name}-gateway-secret"`?  Feels a bit verbose, but less likely to collide with others (I could see someone writing the istio-gateway charm expecting they could own something called `istio-gw-secret`

charms/istio-gateway/metadata.yaml
phvalguima: Okay, I will remove it and retry

charms/istio-pilot/src/charm.py
phvalguima: Agreed, preparing a new commit and will test now with these changes

charms/istio-pilot/src/gateway.yaml.j2
phvalguima: Agreed, I will test it now

charms/istio-gateway/metadata.yaml
phvalguima: I retried without ```series: [kubernetes]``` and it errors with:
```
$ juju deploy ./istio.yaml --trust
Located charm "dex-auth" in charm-hub, channel 2.28/stable
ERROR cannot deploy bundle: min-juju-version in metadata v2 not valid
```
Bundle used: https://pastebin.canonical.com/p/kw7nVtQGJS/

I will keep that setting.

phvalguima: I retried without ```series: [kubernetes]``` and it errors with:
```
$ juju deploy ./istio.yaml --trust
Located charm "dex-auth" in charm-hub, channel 2.28/stable
ERROR cannot deploy bundle: min-juju-version in metadata v2 not valid
```
Bundle used: https://pastebin.canonical.com/p/kw7nVtQGJS/

I will keep that setting.

charms/istio-pilot/src/charm.py
```diff
@@ -138,16 +138,48 @@ def handle_default_gateway(self, event):
...
+            f"{self.app.name}-gateway-secret"
+            if self.model.config["ssl-crt"] and self.model.config["ssl-key"]
+            else None
+        )
+        manifest = None
+        if secret_name:
+            manifest = t.render(
+                name=gateway,
+                secret_name=secret_name,
+                ssl_crt=self.model.config["ssl-crt"],
```
DnPlas, line 162: Do we need `ssl_crt` and `ssl_key` for rendering `gateway.yaml.j2`? If so, they are missing in the yaml file, if not, we can ditch them here.

charms/istio-pilot/src/charm.py
```diff
@@ -138,16 +138,48 @@ def handle_default_gateway(self, event):
...
+            secret = self.env.get_template('gateway-secret.yaml.j2')
+            manifest_secret = secret.render(
+                secret_name=secret_name,
+                ssl_crt=self.model.config["ssl-crt"],
+                ssl_key=self.model.config["ssl-key"],
+                model_name=self.model.name,
+                app_name=self.app.name,
+            )
+            self._apply_manifest(manifest_secret)
+            return
```
DnPlas, line 176: We cannot `return` here as we have a call to `self.handle_ingress(event)` before the method ends. We could leave the part where we define `manifest` out of the `if secret_name:` statement. Something like this:
```python
...
if secret_name:
            secret = self.env.get_template('gateway-secret.yaml.j2')
            manifest_secret = secret.render(
                secret_name=secret_name,
                ssl_crt=self.model.config["ssl-crt"],
                ssl_key=self.model.config["ssl-key"],
                model_name=self.model.name,
                app_name=self.app.name,
            )
            self._apply_manifest(manifest_secret)
manifest = t.render(
            name=gateway,
            secret_name=secret_name,
            model_name=self.model.name,
            app_name=self.app.name,
        )
...
```

charms/istio-pilot/src/charm.py
```diff
@@ -321,7 +353,7 @@ def _get_gateway_address(self):
         """
         # FIXME: service name is hardcoded
         svcs = self.lightkube_client.get(
-            Service, name="istio-ingressgateway", namespace=self.model.name
+            Service, name="istio-ingressgateway-workload", namespace=self.model.name
```
DnPlas, line 356: This is already in the latest charm code, please rebase your branch and omit this change.

charms/istio-pilot/tox.ini
```diff
@@ -21,4 +21,4 @@ paths = {toxinidir}/src {toxinidir}/tests
 [testenv:lint]
 commands =
     flake8 {[vars]paths}
-    black --check {[vars]paths}
+    black --check --diff {[vars]paths}
```
DnPlas, line 24: This PR is for adding SSL support, please push another PR for adding changes to `tox.ini` and related.

DnPlas (commented):
Hi @phvalguima , thanks for this PR. I've left some comments. Also, please rebase your branch to fix the conflicts.

charms/istio-gateway/metadata.yaml
DnPlas: It's very likely that you were building istio from source on a HEAD that had [this](https://github.com/canonical/istio-operators/commit/e8285038956a38bc4dd1de666ba001e6052e4ef4) commit. At the moment we didn't know we had to use `assumes` instead of `min-juju-version`. I suggest you rebase your branch, leave `metadata.yaml` as it is right now in master and try building and deploying istio again. If the problem persists, please let us know, but this shouldn't happen.

DnPlas:
Since @phvalguima is OOO, I am closing this PR as I've opened #99 to track this and add changes as required.

PR #90, opened 2022-05-12
Fixes for missing svc/istio-ingressgateway-workload

DnPlas:
`istio-gateway` has a strong dependency on `istio-pilot`'s relation data to actually create the k8s resources both of the charms need. This PR introduces changes that will ensure the data is sent, as well as the correct messaging and status is shown.


DnPlas:
Due to some errors with the CI, I'm including more commits in this PR to fix them.

PR #110, opened 2022-08-29
Temporal fixes for sending gateway information through relations 

DnPlas:
  * fix: add a check for gateway relation to avoid sending data when units haven't joined  
  * fix: bind handle_gateway_relation to update_status and handle_gateway_relation
    
    When deployed in a bundle with a large number of charms,
    it seems like the istio-pilot charm cannot handle all the
    relations correctly at the same time. Binding handle_gateway_relation
    to update_status ensures that at least gateway_data is sent eventually.

### DIFF
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -158,10 +158,13 @@ class Operator(CharmBase):
         # Update the ingress resources as they rely on the default_gateway
         self.handle_ingress(event)
 
-        # check if gateway is created
-        self.handle_gateway_relation(event)
-
     def handle_gateway_relation(self, event):
+        if self.model.relations["gateway"]:
+            self.log.info(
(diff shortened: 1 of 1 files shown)


charms/istio-pilot/src/charm.py
```diff
@@ -156,10 +158,13 @@ def handle_default_gateway(self, event):
         # Update the ingress resources as they rely on the default_gateway
         self.handle_ingress(event)
 
-        # check if gateway is created
-        self.handle_gateway_relation(event)
-
     def handle_gateway_relation(self, event):
+        if self.model.relations["gateway"]:
+            self.log.info(
+                "No gateway relation found, waiting for remote unit to join the relation"
```
agathanatasha, line 164: How come gateway is not found? Isn't the if statement checking for the relation's existence?

DnPlas:
Closing this PR as it is not based in `main`.

PR #111, opened 2022-08-29
 Temporal fixes for sending gateway information through relations

DnPlas:
  * fix: add a check for gateway relation to avoid sending data when units haven't joined  
  * fix: bind handle_gateway_relation to update_status and handle_gateway_relation
    
    When deployed in a bundle with a large number of charms,
    it seems like the istio-pilot charm cannot handle all the
    relations correctly at the same time. Binding handle_gateway_relation
    to update_status ensures that at least gateway_data is sent eventually.

### DIFF
--- a/charms/istio-pilot/tests/unit/test_charm.py
+++ b/charms/istio-pilot/tests/unit/test_charm.py
@@ -361,17 +361,19 @@ def test_correct_data_in_gateway_relation(harness, mocker, mocked_client):
     )
     mocked_validate_gateway.return_value = True
 
-    rel_id = harness.add_relation("gateway", "app")
-    harness.add_relation_unit(rel_id, "app/0")
     harness.set_model_name("test-model")
     mocker.patch('resources_handler.load_in_cluster_generic_resources')
-    harness.begin_with_initial_hooks()
 
-    gateway_relations = harness.model.relations["gateway"]
-    istio_relation_data = gateway_relations[0].data[harness.model.app]
+    rel_id = harness.add_relation("gateway", "app")
+    harness.add_relation_unit(rel_id, "app/0")
+
+    harness.begin_with_initial_hooks()
+    model = harness.charm.framework.model
+    gateway_relations = model.get_relation("gateway", rel_id)
+    istio_relation_data = gateway_relations.data[harness.charm.app]
 
     assert istio_relation_data["gateway_name"] == harness.model.config["default-gateway"]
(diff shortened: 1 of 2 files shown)


charms/istio-pilot/tests/unit/test_charm.py
```diff
@@ -361,17 +361,19 @@ def test_correct_data_in_gateway_relation(harness, mocker, mocked_client):
...
     mocker.patch('resources_handler.load_in_cluster_generic_resources')
-    harness.begin_with_initial_hooks()
 
-    gateway_relations = harness.model.relations["gateway"]
-    istio_relation_data = gateway_relations[0].data[harness.model.app]
+    rel_id = harness.add_relation("gateway", "app")
+    harness.add_relation_unit(rel_id, "app/0")
+
+    harness.begin_with_initial_hooks()
+    model = harness.charm.framework.model
```
agathanatasha, line 371: what's the difference between `harness.model` and `harness.charm.framework.model`?

DnPlas: The level of abstration, really. I could use `harness.model`.

ca-scribner (approved):
lgtm because this does provide a patch around an ongoing issue, but when merging this can we also open an issue that outlines how we need to fix this better (maybe elaborate the `FIXME` here as a github issue and link it here)?  That way we fix this in future

DnPlas:
> lgtm because this does provide a patch around an ongoing issue, but when merging this can we also open an issue that outlines how we need to fix this better (maybe elaborate the `FIXME` here as a github issue and link it here)? That way we fix this in future

Will [this](https://github.com/canonical/istio-operators/pull/111/files#diff-d3a4e684bd1a3a1822df5ab7605cfce00c97cd7b52f44ed770526c4f4f80eacdR57) `FIXME` suffice?

ca-scribner:
> Will [this](https://github.com/canonical/istio-operators/pull/111/files#diff-d3a4e684bd1a3a1822df5ab7605cfce00c97cd7b52f44ed770526c4f4f80eacdR57) `FIXME` suffice?

I wont block release here, but imo this feels like a real github issue (that we track as a bug and plan to fix).  If someone has a gateway not deploy on them (for example, they notice the problem while waiting for an update-status) there should be a github issue they can find that explains it.  They wont find this code snipped

DnPlas:
> 

Oh, [gotcha](https://github.com/canonical/istio-operators/issues/113). Thansk @ca-scribner

Issue #117, opened 2022-09-07
istio-pilot cannot get gateway ip when deployed with Charmed Kubeflow on AWS


hychiu:
**Environment**

Bundle - Charmed kubeflow 1.6/beta
Platform - OpenShift, 4.9.0-0.okd-2022-02-12-140851
Cloud provider - AWS

**Description**

My `istio-pilot` pod's shows:

```
2022-09-06 08:42:35 INFO juju.worker.uniter.operation runhook.go:146 ran "ingress-relation-changed" hook (via hook dispatching script: dispatch)
2022-09-06 08:42:38 INFO juju-log ingress-auth:10: No gateway address returned - this may be transitory, but if it persists it is likely an unexpected error. Deferring this event 
```

And I found [this line](https://github.com/canonical/istio-operators/blob/main/charms/istio-pilot/src/charm.py#L329) try to get the `ip` field that does not exist on AWS. 

```
$ kubectl get svc istio-ingressgateway-workload -n kubeflow -o yaml
apiVersion: v1
kind: Service
metadata:
  creationTimestamp: "2022-09-06T08:42:34Z"
  finalizers:
  - service.kubernetes.io/load-balancer-cleanup
  labels:
    app: istio-ingressgateway
    app.juju.is/created-by: istio-ingressgateway
    install.operator.istio.io/owning-resource: unknown
    istio: ingressgateway
    istio.io/rev: default
    operator.istio.io/component: IngressGateways
    release: istio
  name: istio-ingressgateway-workload
  namespace: kubeflow
  resourceVersion: "1039086"
  uid: ae2b8c45-94c2-4900-99b5-162efcc7243e
spec:
  allocateLoadBalancerNodePorts: true
  clusterIP: 172.30.134.191
  clusterIPs:
  - 172.30.134.191
  externalTrafficPolicy: Cluster
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - name: http2
    nodePort: 32693
    port: 80
    protocol: TCP
    targetPort: 8080
  - name: https
    nodePort: 30489
    port: 443
    protocol: TCP
    targetPort: 8443
  selector:
    istio: ingressgateway
  sessionAffinity: None
  type: LoadBalancer
status:
  loadBalancer:
    ingress:
    - hostname: xxxxxxxxxxxxxxxxxx.us-east-1.elb.amazonaws.com
``` 

```
[...]

PR #124, opened 2022-10-13
Fix istio gateway address

i-chvets:
**PR for fix of istio gateway address parsing**
Issue reported in: #117 
Work tracked in: #117 

**Summary of changes**
  - Modified code to properly parse Service resource and return either IP or hostname as gateway address.
  - Added unit tests (with disabling of autouse fixture).
  For more details refer to commit logs.

**Outcomes**
`hostname` should be properly parsed when deployed on AWS.

**Testing**
Unit tests are testing for parsing functionality. To test end-to-end deploy on OpenShift/AWS as described in #117

### DIFF
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -63,8 +63,6 @@ jobs:
         # TODO: Pinned to <1.25 until we update to istio 1.15
         channel: '1.24/stable'
         addons: '["dns", "storage", "rbac", "metallb:10.64.140.43-10.64.140.49"]'
-        # Pinned until this bug is resolved: https://bugs.launchpad.net/juju/+bug/1992833
-        bootstrap-options: "--agent-version=2.9.34"
 
     - name: Install dependencies
       run: |
@@ -84,7 +82,8 @@ jobs:
 
     - name: Bootstrap
       run: |
-        sg microk8s -c 'juju bootstrap microk8s uk8s'
+        # Pinned until this bug is resolved: https://bugs.launchpad.net/juju/+bug/1992833
+        sg microk8s -c 'juju bootstrap microk8s uk8s --agent-version=2.9.34'
         sg microk8s -c 'juju add-model istio-system'
 
     - run: sg microk8s -c 'KUBECONFIG=/home/runner/.kube/config tox -e integration
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -23,7 +23,7 @@ jobs:
       with:
         fetch-depth: 0
     - name: Check libs
-      uses: canonical/charming-actions/check-libraries@2.1.0
+      uses: canonical/charming-actions/check-libraries@2.1.1
       with:
         charm-path: ./charms/${{ matrix.charm }}
         credentials: "${{ secrets.charmcraft-credentials }}"
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -60,7 +60,8 @@ jobs:
 
     - uses: balchua/microk8s-actions@v0.2.2
       with:
-        channel: 'latest/stable'
+        # TODO: Pinned to <1.25 until we update to istio 1.15
+        channel: '1.24/stable'
         addons: '["dns", "storage", "rbac", "metallb:10.64.140.43-10.64.140.49"]'
 
     - name: Install dependencies
@@ -86,7 +87,7 @@ jobs:
 
     - run: sg microk8s -c 'KUBECONFIG=/home/runner/.kube/config tox -e integration
         -- --model istio-system --destructive-mode'
-      timeout-minutes: 25
+      timeout-minutes: 40
 
     - name: Setup Debug Artifact Collection
       run: mkdir tmp
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -23,7 +23,7 @@ jobs:
       with:
         fetch-depth: 0
     - name: Check libs
-      uses: canonical/charming-actions/check-libraries@2.1.0-rc
+      uses: canonical/charming-actions/check-libraries@2.1.0
       with:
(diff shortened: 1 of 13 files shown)


.github/workflows/integrate.yaml
```diff
@@ -23,7 +23,7 @@ jobs:
       with:
         fetch-depth: 0
     - name: Check libs
-      uses: canonical/charming-actions/check-libraries@2.0.0-rc
+      uses: canonical/charming-actions/check-libraries@2.1.0
```
ca-scribner, line 26:
```suggestion
      uses: canonical/charming-actions/check-libraries@2.1.1
```

.github/workflows/publish.yaml
```diff
@@ -58,7 +58,7 @@ jobs:
           ref: ${{ inputs.source_branch }}
 
       - name: Select charmhub channel
-        uses: canonical/charming-actions/channel@2.0.0-rc
+        uses: canonical/charming-actions/channel@2.1.0
```
ca-scribner, line 61:
```suggestion
      uses: canonical/charming-actions/check-libraries@2.1.1
```

ca-scribner, line 61:
```suggestion
        uses: canonical/charming-actions/channel@2.1.1
```

.github/workflows/release.yaml
```diff
@@ -21,7 +21,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Release charm to channel
-        uses: canonical/charming-actions/release-charm@2.0.0-rc
+        uses: canonical/charming-actions/release-charm@2.1.0
```
ca-scribner, line 24:
```suggestion
        uses: canonical/charming-actions/release-charm@2.1.1
```

charms/istio-pilot/tests/unit/test_charm.py
```diff
@@ -477,3 +477,54 @@ def test_remove_exceptions(harness, mocked_client, mocker):
...
+
+
+#
+# Test Service
+#
+# Overriding mocked_get(autouse=True) fixture (disabling autouse) to allow setting
+# of return value of get() method.
+@pytest.fixture(scope="function")
+def mocked_get(mocked_charm_client, mocker):
+    return
```
ca-scribner, line 489: I don't understand why this is needed.  I'm also not sure how this affects the fixture.  Does this mean only usages of this fixture later in this file are affect?  Or is there some precedence order?

i-chvets: This is to override `mocked_get(autouse=True)`.
I don't want `mocked_get()` to be patched for my particular test. However, it **is** patched with `autouse=True`. There is no way to disable `autouse` for just one test. As a result, this solution.

ca-scribner: An alternative to this could be to add a `yield mocked_charm_client` to the original `mocked_get` fixture, that way you can just edit its return here.  Although I don't know if that means you're editing the return for all uses, or just for this particular use...

ca-scribner: Actually, I see below you're editing the `.get.return_value` anyway - was that to avoid corrupting the global object?

ca-scribner: But by redefining it here, does it now make `mocked_get` not be an auto-used fixture for the other tests?  I worry that we might be disabling it for something else without realising

i-chvets: > Actually, I see below you're editing the `.get.return_value` anyway - was that to avoid corrupting the global object?
Yes.
My concern is exactly what you described: I don't want to mess with other tests. That's why I've put `scope="function") for `mocked_get()` override.

ca-scribner: If this does override the autouse for all tests, then I think we'll accidentally break tests by doing this (otherwise, why would autouse be set for this fixture at all).  Thinking on this though, I mixed some things up - `autouse` and `scope` are different.  The autoused fixture is `scope=function` (that is the default), which means each test function call gets its own separate instance of the fixture.  So there should not be any bleed across tests.  You should be able to remove the override of mocked_get and leave your other code as is and it will have your desired effect.

ca-scribner: This might be a different problem if the scope was module or other, but the default is function so I don't think you need to work around things here

i-chvets: Without override it does not allow me to set `harness.charm.lightkube_client.get.return_value`. Basically, without override, if I set `harness.charm.lightkube_client.get.return_value` to anything, then access property `_gateway` that uses `get()`, I always get `None`. Override solves this problem.

i-chvets: Changed test to use `side_effect` and removed override of `mocked_get()`

charms/istio-pilot/tests/unit/test_charm.py
```diff
@@ -477,3 +477,50 @@ def test_remove_exceptions(harness, mocked_client, mocker):
     mocked_client.return_value.delete.side_effect = api_error
     with pytest.raises(ApiError):
         harness.charm.on.remove.emit()
+
+
+#
+# Test Service
+#
+def test_service(harness, subprocess, mocked_client, helpers, mocker, mocked_list):
```
ca-scribner, line 485:
```suggestion
def test_service(harness, subprocess, mocked_client, helpers, mocker, mocked_list):
    """Test that the charm._gateway_address correctly returns gateway service IP/hostname."""
```

ca-scribner (changes requested):
one minor suggestion, otherwise lgtm

charms/istio-pilot/tests/unit/test_charm.py
i-chvets: Done

ca-scribner (approved):
lgtm!

PR #132, opened 2022-12-07
feat: integrate istio-pilot with prometheus and grafana

DnPlas:
This PR adds the following:
* [feat: add grafana and prometheus for monitoring](https://github.com/canonical/istio-operators/commit/94e8d169b8d452540968c396521a5afda660c68a)
* [test: add integration test for prometheus, grafana and istio-pilot](https://github.com/canonical/istio-operators/commit/044349e1e39f07c7f9e5bc227c8780178d76b0b9)

NOTE: integration tests may fail due to issues with the current CI, #131 aims at fixing the majority of them.

PR #133, opened 2022-12-07
feat: add alert rules for istio-pilot

DnPlas:
Adds alert rules and corresponding unit/integration tests.

Dear reviewers, this PR depends on #132, and thus should be merged after rebasing on top of a HEAD with the change from that other PR.


i-chvets (approved):
Approved.

PR #151, opened 2022-12-08
Pin runners for ubuntu 20-04, pin renovate python to 3.8.0

misohu:
Pin `runs-on` to `ubuntu-20.04` in every action. `latest` now points to `22.04` and renovate python to `3.8.0`


i-chvets (approved):
Approved.

PR #169, opened 2023-01-31
feat: istioctl version bump 1.11->1.16

DnPlas:
Bump the version of the binary we use in istio-pilot.

#### Relevant manual testing:

* juju deploy local build goes active and idle
* deploy local build of `istio-pilot` alongside the [Charmed Kubeflow 1.6 bundle](https://github.com/canonical/bundle-kubeflow/blob/main/releases/1.6/stable/kubeflow/bundle.yaml), all charms that relate to istio-pilot are active and idle
* after proper configuration, the Kubeflow Dashboard (from provided by the bundle ^) can be accessed


i-chvets (approved):
Approed.

PR #171, opened 2023-02-03
ci: multiple fixes for track/1.11 CI

DnPlas:

* ci: pin ubuntu 20.04 in integrate, publish, and release jobs
Pinning the version avoids incompatibilities with building the charm and installing certain dependencies.
* build: pin lightkube and lightkube-models to versions that are compatible with the charm and test code
* bootstrap juju with agent 2.9.34 to avoid https://bugs.launchpad.net/juju/+bug/1992833

### DIFF
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -81,7 +81,7 @@ jobs:
 
     - name: Bootstrap
       run: |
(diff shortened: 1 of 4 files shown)


i-chvets (approved):
Approved.

PR #177, opened 2023-02-22
feat: unpin juju agent-version

ca-scribner:
juju `agent-version` was previously pinned to 2.9.34 due to https://bugs.launchpad.net/juju/+bug/1992833.  This has been resolved.

PR #178, opened 2023-02-23
build: pin pythonlib-juju to <3.1 to avoid incompatibilty issues with juju 2.9.x

DnPlas:
To avoid issues where 2.9.x is not suported. Example:

```
File "/home/runner/work/seldon-core-operator/seldon-core-operator/.tox/integration/lib/python3.8/site-packages/juju/client/connector.py", line 82, in connect
    raise JujuConnectionError("juju server-version %s not supported" % juju_server_version)
juju.errors.JujuConnectionError: juju server-version 2.9.34 not supported
```
In this PR you will find changes in `requirements-integration.in`, where we are pinning the pythonlib-juju version and in `requirements-integration.txt` which is an auto-generated file based on the dependencies of the aforementioned lib.

Merge before #177


ca-scribner (changes requested):
I like the idea, but I think you need to render the pip-compile using python 3.8.  Otherwise, we get the wrong websockets version (and maybe this appnope thing?)

ca-scribner (approved):

PR #176, opened 2023-02-21
feat: add option to use `NodePort` for istio-gateway

ca-scribner:
## Context

#175 and elsewhere requests that the `istio-gateway` charm allow for access through a `NodePort` rather than just through a `LoadBalancer` or `ClusterIP`.  This feature is implemented here, where add another valid entry for `gateway_service_type`, `gateway_service_type=NodePort`

Because of how `istio-pilot` was using the service provided by `istio-gateway`, we also need to change some of the logic around how `istio-pilot` interrogates the gateway it finds.  Included here too are these changes, as well as an expanded test suite.

See #175 for more context.  closes #175

## Testing instructions

Note that these instructions use charmhub `edge/branch` charms.  If you don't have access to the charmhub branch charms, you must build istio-pilot and istio-gateway for yourself locally.

(edited Feb 22 to fix the auth config)

```
juju add-model kubeflow

# Deploy `istio-pilot` and `istio-gateway`, where `istio-gateway` uses the `gateway_service_type=NodePort`
juju deploy istio-pilot --channel=edge/add-gateway-nodeport --trust --series kubernetes
juju deploy istio-gateway --channel=edge/add-gateway-nodeport --trust --config kind=ingress --config gateway_service_type=NodePort istio-ingressgateway --series kubernetes
juju relate istio-pilot istio-gateway
# Charms should go to Active

# Look for the gateway service, which should be of type nodeport:
kubectl get svc -n kubeflow | grep istio-ingressgateway-workload
# should see something like:
# istio-ingressgateway-workload    NodePort    10.152.183.121   <none>        80:32693/TCP,443:30358/TCP              23m
# Where the "80:XXXXX/TCP" shows what port is open for regular http traffic (used below as NODEPORT)


# Deploy some things that we can access through the NodePort to confirm it works
juju deploy kubeflow-dashboard --channel edge --trust
juju deploy kubeflow-profiles --channel edge --trust
juju relate kubeflow-dashboard kubeflow-profiles
juju relate kubeflow-dashboard istio-pilot
juju relate kubeflow-dashboard:ingress istio-pilot:ingress
# Charms should go to Active

juju deploy kubeflow-volumes --channel 1.6/stable
juju relate kubeflow-volumes:ingress istio-pilot:ingress
# Charms should go to Active

# Test access via nodeport
# In a browser, browse to `YOUR_LOCAL_IP:NODEPORT`.  You should be able to access the dashboard (likely after creating a user profile)
[...]

PR #179, opened 2023-02-23
feat: Backport `NodePort` to track 1.11

ca-scribner:
This is a backport of the NodePort feature implemented in the main branch for PR #176.  This attempts to extract any non-relevant features from that branch, importing only those needed to support NodePort gateways.  See #176 for more details on functionality.

(cherry picked from commit 0dd331421462a539b427fbd95d1bdb91d57311ce, but with some modifications)

### DIFF
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -12,8 +12,14 @@ from ops.charm import CharmBase, RelationBrokenEvent
 from ops.main import main
 from ops.model import ActiveStatus, BlockedStatus, WaitingStatus
 from serialized_data_interface import NoCompatibleVersions, NoVersionsListed, get_interfaces
+
+from istio_gateway_info_provider import RELATION_NAME, GatewayProvider
 from resources_handler import ResourceHandler
-from istio_gateway_info_provider import GatewayProvider, RELATION_NAME
+
+GATEWAY_HTTP_PORT = 8080
+GATEWAY_HTTPS_PORT = 8443
+METRICS_PORT = 15014
+GATEWAY_WORKLOAD_SERVICE_NAME = "istio-ingressgateway-workload"
 
 
 class Operator(CharmBase):
@@ -60,11 +66,11 @@ class Operator(CharmBase):
         for event in [self.on[RELATION_NAME].relation_changed, self.on.update_status]:
             self.framework.observe(event, self.handle_gateway_info_relation)
         self.framework.observe(self.on["istio-pilot"].relation_changed, self.send_info)
-        self.framework.observe(self.on['ingress'].relation_changed, self.handle_ingress)
-        self.framework.observe(self.on['ingress'].relation_broken, self.handle_ingress)
-        self.framework.observe(self.on['ingress'].relation_departed, self.handle_ingress)
-        self.framework.observe(self.on['ingress-auth'].relation_changed, self.handle_ingress_auth)
-        self.framework.observe(self.on['ingress-auth'].relation_departed, self.handle_ingress_auth)
+        self.framework.observe(self.on["ingress"].relation_changed, self.handle_ingress)
+        self.framework.observe(self.on["ingress"].relation_broken, self.handle_ingress)
+        self.framework.observe(self.on["ingress"].relation_departed, self.handle_ingress)
+        self.framework.observe(self.on["ingress-auth"].relation_changed, self.handle_ingress_auth)
+        self.framework.observe(self.on["ingress-auth"].relation_departed, self.handle_ingress_auth)
 
     def install(self, event):
         """Install charm."""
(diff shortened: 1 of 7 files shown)


ca-scribner:
I have tested this using the same instructions as in #176, except that I had to build the charms locally as I think pulling from a charmhub branch that is not on the default track is bugged?  Otherwise, it worked the same as #176 did

charms/istio-pilot/tests/unit/test_charm.py
```diff
@@ -29,17 +33,17 @@ def side_effect(*args, **kwargs):
...
 
     harness.set_leader(True)
     harness.begin_with_initial_hooks()
 
-    install = mocker.patch('charm.Operator.install')
-    remove = mocker.patch('charm.Operator.remove')
-    send_info = mocker.patch('charm.Operator.send_info')
-    handle_ingress = mocker.patch('charm.Operator.handle_ingress')
-    handle_ingress_auth = mocker.patch('charm.Operator.handle_ingress_auth')
+    install = mocker.patch("charm.Operator.install")
```
DnPlas, line 42: did you run black on the charm code or is this introduced by the backported charm code?

charms/istio-pilot/src/charm.py
```diff
@@ -286,64 +294,126 @@ def handle_ingress_auth(self, event):
...
             return
 
-        t = self.env.get_template('auth_filter.yaml.j2')
-        auth_filters = ''.join(
+        if self.model.config["ssl-crt"] and self.model.config["ssl-key"]:
+            gateway_port = GATEWAY_HTTPS_PORT
+        else:
+            gateway_port = GATEWAY_HTTP_PORT
+
+        t = self.env.get_template("auth_filter.yaml.j2")
```
DnPlas, line 302: Seems like this comes from another commit. I believe we merged this only in main, but not in 1.11 (?)

charms/istio-pilot/src/charm.py
```diff
@@ -286,64 +294,126 @@ def handle_ingress_auth(self, event):
...
+            exporter_service = self.lightkube_client.get(
+                res=Service, name="istiod", namespace=self.model.name
+            )
+            exporter_ip = exporter_service.spec.clusterIP
+        except ApiError as e:
+            if e.status.code == 404:
+                return None
+            raise
+        else:
+            return exporter_ip
```
DnPlas, line 344: This was also introduced by another commit, I believe this comes from the observability integration.

DnPlas (commented):
Thanks @ca-scribner, I noticed this PR is introducing a couple of things that were merged in `main`, but never backported into `track/1.11`. I spotted the observability integration and the ssl enablement. I think it would be much healthier for 1.11 if we only backported just the `NodePort` changes.

ca-scribner:
> by "bugged" you mean you cannot deploy from 1.11/edge/backport-nodeport-to-track-1.11 ?

@DnPlas correct.  If I do this, it has trouble pulling images for some reason

charms/istio-pilot/tests/unit/test_charm.py
ca-scribner: Yeah you mean because of the changes in `'` to `"`?  somewhere between when this branch departed from main and now, `black` has been run on main, so these quote changes are coming from main via the backport.  They aren't necessary, but I took them in mainly so that I didn't introduce an error in the backporting.  They can be removed

charms/istio-pilot/src/charm.py
ca-scribner: Removed

charms/istio-pilot/src/charm.py
ca-scribner: You're right, this was mangled into the cherry picked code.  I have removed it now

charms/istio-pilot/tests/unit/test_charm.py
DnPlas: Yeh, I just wanted to know as I was suspecting some other changes were introduced. We can keep them as long as it is just the formatting.

charms/istio-pilot/src/charm.py
```diff
@@ -286,64 +294,106 @@ def handle_ingress_auth(self, event):
...
-        """
-        # FIXME: service name is hardcoded
+    def _is_gateway_service_up(self):
+        """Returns True if the ingress gateway service is up, else False."""
+        svc = self._get_gateway_service()
+
+        if svc.spec.type == "NodePort":
+            # TODO: do we need to interrogate this further for status?
+            return True
+        if _get_gateway_address_from_svc(svc) is not None:
```
misohu, line 333: `_get_gateway_address_from_svc` returns address from service .... If I am not wrong we never use the address so I am just curious if its finding the address of the services worth the hustle. If not we can just remove the whole `_get_address_from_loadbalancer` and return true if the service is there. Or is there a chance the service is up without address ? if so dont we also need to check some more info about the nodeport service ? (e.g. if port was really assigned).

misohu (approved):
LGTM just one comment with question.

charms/istio-pilot/src/charm.py
ca-scribner: You're right, we never use the address as an address.  It is a surrogate for whether the LoadBalancer is up and functioning like you mention.  That the service exists is not enough (for example, think of a LoadBalancer service in microk8s when we haven't deployed metallb - it is a service, but it does not have an ip).  It doesn't feel like the best approach, but I don't think we've come up with something better.

For the NodePort, I agree that it would be better to interrogate it further, but I didn't see anything more available.  It doesn't have a status or other attributes that indicated anything.  I don't think it can be up without a port, but I'm not sure

PR #191, opened 2023-03-14
feat: istioctl wrapper

ca-scribner:
To improve how we use istioctl and to enable better error handling, this PR implements an Istioctl wrapper class that operates the istioctl tool.  The intent here is to stop calling istioctl directly in our charm, instead always using this wrapper class.  

### Notes for testers:

* Please try to actually use this Istioctl class to create/upgrade/remove istio in your own local microk8s.  You can just start a python session, `from istioctl import Istioctl`, and start managing your istio.  See the tests for some ideas on how to use it.


charms/istio-pilot/src/istioctl.py
```diff
@@ -0,0 +1,214 @@
...
+class UpgradeFailedError(Exception):
+    pass
+
+
+class VersionCheckError(Exception):
+    pass
+
+
+class Istioctl:
+    def __init__(
```
i-chvets, line 30: Add doc string

charms/istio-pilot/src/istioctl.py
```diff
@@ -0,0 +1,214 @@
...
+        TODO: Should we use `istioctl x uninstall` here instead of lightkube?  It is an
+        experimental feature but included in all istioctl versions we support.
+        """
+        manifest = self.manifest()
+
+        # Render YAML into Lightkube Objects
+        k8s_objects = codecs.load_all_yaml(manifest, create_resources_for_crds=True)
+
+        client = Client()
+        delete_many(client=client, objs=k8s_objects)
```
i-chvets, line 117: Do we want exception handling around this?
I.e. if status is not 404, we log/raise error.

i-chvets:
Testing session, not sure if this should have failed precheck:
```
>>> from istioctl import Istioctl
>>> ictl = Istioctl(istioctl_path="../istio-1.16.2/bin/istioctl", namespace="kf", profile=default)
  File "<stdin>", line 1, in <module>
NameError: name 'default' is not defined
>>> ictl = Istioctl(istioctl_path="../istio-1.16.2/bin/istioctl", namespace="kf", profile="default")
>>> ictl.precheck()
✔ No issues found when checking the cluster. Istio is safe to install or upgrade!
  To get started, check out https://istio.io/latest/docs/setup/getting-started/
>>> ictl.upgrade()
✔ No issues found when checking the cluster. Istio is safe to install or upgrade!
  To get started, check out https://istio.io/latest/docs/setup/getting-started/
WARNING: Istio control planes installed: 1.11.0.
WARNING: A newer installed version of Istio has been detected. Running this command will overwrite it.
✔ Istio core installed                                                                                                                                         
✔ Istiod installed                                                                                                                                             
✔ Ingress gateways installed                                                                                                                                   
✔ Installation complete                                                                                                                                        Making this installation the default for injection and validation.
>>> 
```

i-chvets:
OK. I figured it out.
Will approve when done with CI.

charms/istio-pilot/src/istioctl.py
```diff
@@ -0,0 +1,214 @@
...
+            raise ManifestFailedError(error_msg) from cpe
+
+        return manifests
+
+    def precheck(self):
+        """Executes `istioctl x precheck` to validate whether the environment can be updated.
+
+        Raises:
+            PrecheckFailedError: if the precheck command fails.
+        """
```
i-chvets, line 91: I think, we need a more info in comments regarding when `precheck` will succeed/fail. `precheck` will succeed on checking 1.11 against 1.16.2. And I understand that that's exactly what Istio precheck does, but since it is our wrapper, we can specify in comments or may be reference Istio doc that explains things.

DnPlas: Yeah I would also like to see more of this, what would a successful precheck be and what does it mean, what useful information can we get from it?

charms/istio-pilot/src/istioctl.py
```diff
@@ -0,0 +1,214 @@
...
+        manifest = self.manifest()
+
+        # Render YAML into Lightkube Objects
+        k8s_objects = codecs.load_all_yaml(manifest, create_resources_for_crds=True)
+
+        client = Client()
+        delete_many(client=client, objs=k8s_objects)
+
+    def upgrade(self, precheck: bool = True):
```
DnPlas, line 119: Would upgrading using this method require us to update the istioctl binary that we set in charmcraft.yaml? 
Also, this seems to me like we don't really need any other charm to upgrade to. How does this method integrate with `juju refresh istio-pilot --channel <channel>`?

ca-scribner: The `Istioctl` class only contains how to operate **an** `istioctl`, nothing specific about a particular version of istioctl.  Upgrading using this method uses whatever istioctl binary you've given it to try to upgrade the k8s cluster.  

For the istio-pilot 1.16.2 charm, we'd include istioctl 1.16.2 in charmcraft.yaml (and similarly for 1.11.0, we'd include istioctl 1.11.0).  For this reason, yes atm we do still need a charm for each istio-pilot version.  We *could* add multiple binaries into the same charm (so istio-pilot 1.16.2 could also have older binaries included), but then we'd get much larger images because of the extra binaries, etc

charms/istio-pilot/src/istioctl.py
```diff
@@ -0,0 +1,214 @@
...
+class PrecheckFailedError(Exception):
+    pass
+
+
+class UpgradeFailedError(Exception):
+    pass
+
+
+class VersionCheckError(Exception):
+    pass
```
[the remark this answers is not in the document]
beliaev-maksim: if you will try except them and do different treatments, then it might be the case. However, so far I do not see any try/except block for these errors :)

charms/istio-pilot/src/istioctl.py
ca-scribner: I think this is resolved now?

charms/istio-pilot/src/istioctl.py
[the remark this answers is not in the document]
beliaev-maksim: but then we use only one of them per function, which means, that there is no branch required. 
So, if you raise for example `UpgradeFailedError` in all of them, nothing will be changed.

charms/istio-pilot/src/istioctl.py
ca-scribner: Agreed this is not good.  Since the charm already had a removal procedure, I've just copied it here.  You're right though that it'll fail.  The failure you posted in the main conversation is also a common thing (this charm rarely, if ever, removes itself correctly).  

Since we already had a remove procedure I didn't want to creep the scope of this task to change it, but I think we should fix the remove procedure too. There's an `istioctl x uninstall --purge` that will probably work better for us, and is more in line with how we integrate with everything else. 

What do you think about us opening this as a separate issue, but trying to resolve it right away?

ca-scribner: ahh damn.  Just tested and `istioctl x uninstall --purge` will also kill `istio-gateway`'s deployment, breaking that charm as well.  So it is not a good candidate for the remove hook here.

charms/istio-pilot/src/istioctl.py
```diff
@@ -0,0 +1,221 @@
...
+            logging.error(error_msg)
+            logging.error(f"stdout: {cpe.stdout}")
+            logging.error(f"stderr: {cpe.stderr}")
+
+            raise ManifestFailedError(error_msg) from cpe
+
+        return manifests
+
+    def precheck(self):
+        """Executes `istioctl x precheck` to validate whether the environment can be updated.
```
i-chvets, line 92:
```suggestion
        """Executes `istioctl x precheck` to validate whether the environment can be updated.

        NOTE: This function does not validate exact versions compatibility. This verification should be done by caller.
```

charms/istio-pilot/src/istioctl.py
i-chvets: Suggested. Just fix my wording, if you'd like.

charms/istio-pilot/src/istioctl.py
i-chvets: Sounds good. Let's open separate remove handling issue. There will be more things in remove handler for sure.

i-chvets:
LGTM.

charms/istio-pilot/src/istioctl.py
ca-scribner: lgtm!  ty

charms/istio-pilot/src/istioctl.py
ca-scribner: I think we've sorted this out offline and this can be resolved, but please reopen if I'm incorrect!

charms/istio-pilot/src/istioctl.py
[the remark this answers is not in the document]
ca-scribner: ahh I see what you're saying.  You're saying that `Istioctl.precheck` only raises one Exception, `istioctl.upgrade` only raises one Exception, etc, so they don't need to be different types.  Yeah that's true

We can collapse them into one, although I think a single custom exception still makes catching it in the charm easier.  Would that work for you?  Maybe just `IstioctlError`?  I'll patch that now

i-chvets (approved):
Approved.

PR #198, opened 2023-03-22
feat: Add upgrade handling to istio-pilot

### LINKED ISSUE #188: istio-pilot not upgrading istiod properly on refresh
istio-pilot is not upgrading the deployed version of istio when the charm is refreshed.  Take the following case:

```
juju add-model kubeflow
juju deploy istio-pilot --channel=1.11/stable --trust
juju deploy istio-gateway --channel=1.11/stable --trust --config kind=ingress istio-ingressgateway

# Wait for things to come up

juju refresh istio-pilot --channel=1.16/edge
```

After the refresh is done, we can check the istio version using [istioctl](https://istio.io/latest/docs/reference/commands/istioctl/):

```
istioctl version -i kubeflow

client version: 1.11.0
control plane version: 1.11.0  <----------------------- Deployed version
data plane version: 1.11.0 (1 proxies)
```

We can also see an issue if we look at `kubectl get pods`
```
kubectl get pods

NAME                                             READY   STATUS    RESTARTS   AGE
istio-ingressgateway-0                           1/1     Running   0          66m
istiod-65789675d4-sm2g4                          1/1     Running   0          65m
istio-ingressgateway-workload-7885758c97-jhtqk   1/1     Running   0          64m
istio-pilot-0                                    1/1     Running   0          33m
```

Note the age on the istio-pilot-0 (charm) pod vs the istiod (istio's daemon) pod - the charm pod is younger than the daemon because the charm refreshed whereas the daemon is the old one.

afaict, refreshing istio-pilot gives us the 1.16 **charm**, but does nothing to istio's own controllers or resources.  

I think this makes sense, because on a refresh/upgrade we will trigger the `upgrade-charm` event, **not** the `install` event.  Looking at istio-pilot we see we deploy resources [only when handling the install event](https://github.com/canonical/istio-operators/blob/420ebfa153f207c767a1a0aab1109580abdb9ce7/charms/istio-pilot/src/charm.py#L85) - we do not handle the `upgrade-charm` event.

If we were to bind our current `install` handler to `upgrade-charm` as well, that might result in a working istio upgrade.  Testing locally by executing the `istioctl_1.16 install` command on my existing istio 1.11 yields:

```
/opt/istioctl/istio-1.16.2/bin/istioctl install -y -s profile=minimal -s values.global.istioNamespace=kubeflow

WARNING: Istio control planes installed: 1.11.0.
WARNING: A newer installed version of Istio has been detected. Running this command will overwrite it.
✔ Istio core installed                                                                                                                                                                   
✔ Istiod installed                                                                                                                                                                       
✔ Installation complete                                                                                                                                                                  Making this installation the default for injection and validation.

Thank you for installing Istio 1.16.  Please take a few minutes to tell us about your install/upgrade experience!  https://forms.gle/99uiMML96AmsXY5d6
```

Now if I check the version I get 1.16 as desired:

```
istioctl version -i kubeflow

client version: 1.11.0
control plane version: 1.16.2
data plane version: 1.11.0 (1 proxies)
```

Istio's prerequisites state that ["the installed istio version is no more than one minor version less than the upgrade version.  For example, 1.6.0 or higher is required before you start the upgrade process to 1.7.x"](https://istio.io/latest/docs/setup/upgrade/in-place/#upgrade-prerequisites), so while this upgrade did not fail, we are not following istio's stated path for upgrades by going straight from 1.11 to 1.16.

ca-scribner:
### Summary

This enables the istio-pilot charm to upgrade Istio by up to one minor version.  This should be merged after #191 as it uses the `Istioctl` wrapper class to execute `istioctl`.  

This is designed to [this spec](https://docs.google.com/document/d/14SLwwW3-hWR5fAGyYWYPOg4ERGqI3CQlDF13MQ52FYw/edit#)

### Known bugs

`istioctl` v1.12, 1.13, and 1.14 have a bug where `istioctl upgrade` incorrectly creates a `ValidatingWebhookConfiguration` `istiod-default-validator` that looks for `istiod` in the default `istio-system` namespace, not the namespace we set (for us, typically `kubeflow`).  This results in a failing webhook immediately after upgrading **to** any of these versions.  This does not affect upgrading to istio v1.15 or 1.16, and will be fixed for 1.12-1.14 in a follow-on PR.

### Out of scope for this PR

* Deploying the other istio-pilot versions (1.12 through 1.15)

### Testing instructions

No automated upgrade tests are included yet, but manual testing is possible.  For example, you can do:
* `charmcraft pack` the istio-pilot and istio-gateway charms here
* `juju deploy` the two charms
* `juju refresh istio-pilot --path ./istio-pilot_ubuntu-20.04-amd64.charm`  # <--this refreshes the charm to **the same** version, which will work as defined in the spec

To test actual upgrades, make copies of your `./istio-pilot_ubuntu-20.04-amd64.charm` charm file and replace the `istioctl` in those copies with different versions (get different versions from [here](https://github.com/istio/istio/releases/)).  For example, I tested with:
* istio-pilot_ubuntu-20.04-amd64_istioctl_1.11.0.charm
* istio-pilot_ubuntu-20.04-amd64_istioctl_1.15.6.charm
* istio-pilot_ubuntu-20.04-amd64_istioctl_1.16.2.charm

With these different versions of the charm, you can now test things like
* deploy 1.15.6 and refresh to 1.16.2 (should work)
* deploy 1.16.2 and refresh to 1.15.2 (should fail), then refresh again back to 1.16.2 and it should work again
* deploy 1.15.6, refresh to 1.11.0 (should fail), then refresh to 1.16.2 (should work)
* deploy to 1.11.0, refresh to 1.16.2 (should fail), then **manually** use istioctl's 1.12 through 1.15 to update istio yourself to 1.15.  Once you're at 1.15, you can `juju resolve istio-pilot` your 1.16.2 deployment, and it will finish the upgrade and start working

Please try to poke many holes, read the error messages, say where things are unclear, etc!

---

Closes #188


charms/istio-pilot/src/charm.py
```diff
@@ -126,6 +139,58 @@ def remove(self, event):
...
+            raise GenericCharmRuntimeError("Failed to upgrade.  See `juju debug-log` for details.") from e
+
+        # TODO: Add an override that can be set by config, eg:
+        #  `juju config istio-pilot skip-upgrade=true`, so we can get out of really sticky
+        #  situations like if our version parsing is wrong54
+
+        # Use istioctl precheck to confirm the upgrade should be safe
+        try:
+            self._log_and_set_status(MaintenanceStatus("Excecuting `istioctl precheck`"))
+            istioctl.precheck()
```
i-chvets, line 170: Here I would refer to my comments in [wrapper class PR](https://github.com/canonical/istio-operators/pull/191):
`precheck` will return True if existing installation is 1.11 and `istioctl` version we are using is 1.16.2
Not sure if this is right behaviour.

charms/istio-pilot/src/charm.py
```diff
@@ -126,6 +139,58 @@ def remove(self, event):
...
+        """
+        istioctl = Istioctl(ISTIOCTL_PATH, self.model.name, ISTIOCTL_DEPOYMENT_PROFILE)
+        self._log_and_set_status(MaintenanceStatus("Upgrading Istio"))
+
+        # Check for version compatability for the upgrade
+        try:
+            versions = istioctl.version()
+        except VersionCheckError as e:
+            self.log.error(UPGRADE_FAILED_MSG.format(message=str(e)))
+            raise GenericCharmRuntimeError("Failed to upgrade.  See `juju debug-log` for details.") from e
```
DnPlas, line 155: I think this line and the previous one can send a different message. IMO, they should say the version of control plane and client could not be retrieved and because of that the upgrade will not proceed. This check didn't even try to upgrade, so it's fair to have a more specific message of what happened.

charms/istio-pilot/src/charm.py
```diff
@@ -426,5 +513,40 @@ def _get_address_from_loadbalancer(svc):
...
+        versions (dict): A dictionary containing:
+                            client: the client version (eg: istioctl)
+                            control_plane: the control plane version (eg: istiod)
+
+    Returns True if this is the case, else raises an exception with details.
+    """
+    client_version = Version(versions["client"])
+    control_plane_version = Version(versions["control_plane"])
+
+    if client_version < control_plane_version:
```
DnPlas, line 532: I am confused about these comparisons. Where do you tell istioctl the version you want to upgrade to so that the client version and the control plane version are different?

charms/istio-pilot/src/charm.py
```diff
@@ -12,22 +11,35 @@
...
+from istioctl import Istioctl, PrecheckFailedError, UpgradeFailedError, VersionCheckError
 from resources_handler import ResourceHandler
 
 GATEWAY_HTTP_PORT = 8080
 GATEWAY_HTTPS_PORT = 8443
 METRICS_PORT = 15014
 GATEWAY_WORKLOAD_SERVICE_NAME = "istio-ingressgateway-workload"
+ISTIOCTL_PATH = "./istioctl"
+ISTIOCTL_DEPOYMENT_PROFILE = "minimal"
+UPGRADE_FAILED_MSG = "Failed to upgrade Istio.  {message}  To recover Istio, use istioctl to " \
```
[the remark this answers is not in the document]
ca-scribner: atm, it is the charm calling the istioctl that was packed into the charm (like previous istio-pilot versions), not what Jon suggested where we include it as a separate image.  There is an istioctl image we could use, but I haven't done that here.  That would be a good follow on task though

charms/istio-pilot/src/charm.py
ca-scribner: istioctl (the binary) only knows how to update to its own version.  So istioctl_1.16.2 knows how to upgrade to 1.16.2, nothing else.  Executing `istioctl version` returns the client (istioctl) and control_plane (whatever is in k8s) versions.  That means `istioctl version` tells us both what we are trying to upgrade to (client) and what we're upgrading from (control_plane)

charms/istio-pilot/requirements-unit.txt
```diff
@@ -23,8 +23,14 @@ chardet==3.0.4
     # via
     #   -r requirements.txt
     #   requests
+charmed-kubeflow-chisme==0.0.5
```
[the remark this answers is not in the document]
ca-scribner: After several minutes of head scratching thinking "didn't I spend a while getting 0.0.8 to work?" I realised I had a commit that didn't get pushed :) 
 Should be updated now

charms/istio-pilot/src/charm.py
```diff
@@ -12,22 +11,35 @@
...
 from ops.charm import CharmBase, RelationBrokenEvent
 from ops.main import main
-from ops.model import ActiveStatus, BlockedStatus, WaitingStatus
+from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingStatus
 from serialized_data_interface import NoCompatibleVersions, NoVersionsListed, get_interfaces
+from packaging.version import Version
+import yaml
 
+# TODO: Use directly from Chisme when the pyyaml version conflict between ops and sdi is solved by
+#  [this pr](https://github.com/canonical/serialized-data-interface/pull/45)
```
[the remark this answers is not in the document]
ca-scribner: Yep good catch - removed it now!

charms/istio-pilot/src/charm.py
ca-scribner: as is if we fail here, we get a a status of:
> Failed to upgrade.  See `juju debug-log` for details.

and in the debug-log we get:
> Failed to upgrade Istio.  Failed to get client version - no version found in output.  To recover Istio, use istioctl to ...

Would the debug-log message be better as: 
> Failed to upgrade Istio because of an error retrieving version information about istio.  Got message: '{message}' when trying to retrieve version information.

?

I wonder if I should add some general "if an upgrade goes wrong, here's some things to try" instructions to the README.md in this repo.  It is probably easier to have general instructions there, and then it also can be updated live.  Thoughts?

ca-scribner: For now, I've changed it to the new message listed, but curious what you think

charms/istio-pilot/src/charm.py
```diff
@@ -376,6 +441,28 @@ def _get_gateway_service(self):
         )
         return svc
 
+    def _log_and_set_status(self, status):
```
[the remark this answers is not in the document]
ca-scribner: Updated things a little more.  I've added to the istio-pilot README.md some general instructions on fixing a failed upgrade, and then I've linked the doc in the error message.  What do you think

charms/istio-pilot/src/charm.py
ca-scribner: Updated things a little more. I've added to the istio-pilot README.md some general instructions on fixing a failed upgrade, and then I've linked the doc in the error message. What do you think?

charms/istio-pilot/src/istioctl.py
```diff
@@ -0,0 +1,201 @@
...
+                [
+                    self._istioctl_path,
+                    "upgrade",
+                    "-y",
+                    *self._istioctl_flags,
+                ]
+            )
+        except subprocess.CalledProcessError as cpe:
+            raise IstioctlError(
+                "Upgrade failed during `istioctl upgrade` with error code" f" {cpe.returncode}"
```
beliaev-maksim, line 129:
```suggestion
                f"Upgrade failed during `istioctl upgrade` with error code {cpe.returncode}"
```

charms/istio-pilot/src/istioctl.py
```diff
@@ -0,0 +1,201 @@
...
+        raise IstioctlError(error_message_template.format(message="too many meshes found"))
+
+    mesh = meshes[0]
+
+    try:
+        if mesh["Component"] != "pilot":
+            raise IstioctlError(error_message_template.format(message="no control plane found"))
+        version = mesh["Info"]["version"]
+    except KeyError:
+        raise IstioctlError(error_message_template.format(message="no control plane found"))
```
beliaev-maksim, line 199: where is the missing key which might cause the exception? If component, then could be simplified to:
```suggestion
    if "Component" not in mesh or mesh["Component"] != "pilot":
        raise IstioctlError(error_message_template.format(message="no control plane found"))
    version = mesh["Info"]["version"]
```

beliaev-maksim (changes requested):

charms/istio-pilot/src/istioctl.py
ca-scribner: Resolved locally and pushed.

charms/istio-pilot/src/istioctl.py
ca-scribner: I don't know if there's cases where this data can be incomplete, so I'm trying to protect against any of these missing keys (`Component`, `Info`, or `Info.version`).  It could be collapsed, but it'll be something like:
```python
if "Component" not in mesh 
  or mesh["Component"] != "pilot" 
  or "Info" not in mesh 
  or "version" not in mesh["Info"]:
  raise IstioctlError(error_message_template.format(message="no control plane found"))
version = mesh["Info"]["version"]
```

charms/istio-pilot/src/charm.py
ca-scribner: I think this is sorted, but following up with @dnplas offline.  Resolving for now

charms/istio-pilot/src/charm.py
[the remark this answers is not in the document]
ca-scribner: I think this is sorted, but following up with @dnplas offline.  Resolving for now

beliaev-maksim (approved):

PR #206, opened 2023-03-27
docs: add detail to istio-pilot upgrade recovery instructions

ca-scribner:
Improves instructions around recovering from failing upgrades, particularly for when users are stuck in an Error state.

### DIFF
--- a/charms/istio-pilot/README.md
+++ b/charms/istio-pilot/README.md
@@ -27,16 +27,25 @@ This charm supports in-place upgrades of the Istio control plane by up to one mi
 
 The charm will attempt to validate the upgrade prior to modifying the control plane, setting an `ErrorStatus` and logging the reason if an unsupported upgrade is detected.
 
-When upgrading istio-pilot, you should also upgrade istio-gateway at the same time to the same version.  These charms are only tested against each other at the same version.
+### Typical Upgrade Procedure
 
-If you need to upgrade across multiple versions (say from Istio 1.11.0 to Istio 1.16.0), refresh the charm one minor version at a time.  For example, starting with a deployed istio-pilot 1.11, you can do:
+When upgrading istio-pilot, you should **always remove the istio-gateway charm before upgrading** (some versions of `istio-pilot` do not upgrade properly if istio-gateway is deployed).  If you are upgrading one minor version, the procedure is:
+* `juju remove-application istio-ingressgateway`
+* `juju refresh istio-pilot --channel <desired-version>/stable`
+* `juju deploy istio-gateway --channel <desired-version>/stable --trust --config kind=ingress istio-ingressgateway`
+
+Where the istio-gateway config should match whatever you were using before upgrading.   
+
+For upgrades across multiple versions (say from Istio 1.11.0 to Istio 1.16.0), use the same procedure as above but refresh istio-pilot multiple times through each intermediate minor version.  For example, starting with a deployed istio-pilot 1.11, you can do:
+* `juju remove-application istio-ingressgateway`
 * `juju refresh istio-pilot --channel 1.12/stable`
 * `juju refresh istio-pilot --channel 1.13/stable`
 * `juju refresh istio-pilot --channel 1.14/stable`
 * `juju refresh istio-pilot --channel 1.15/stable`
 * `juju refresh istio-pilot --channel 1.16/stable`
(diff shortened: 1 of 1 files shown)


charms/istio-pilot/README.md
```diff
@@ -48,9 +57,13 @@ If at any point you use the [istioctl](https://istio.io/latest/docs/reference/co
...
+When the charm detects an upgrade across more than one minor version, it will set an `ErrorStatus` and log the reason.  For example, if you have istio-pilot 1.13.0 deployed and you `juju refresh istio-pilot --channel 1.15/stable`, the upgrade-charm event will fail and the charm will go to `ErrorStatus`.  To resolve this error:
+* refresh the charm again back to a supported version (`juju refresh istio-pilot --channel 1.14/stable`)
+* if this refresh does not clear the error state immediately, use `juju resolve istio-pilot/0 --no-retry` to clear the error on the previous refresh event and move on to the next event.  This should lead to the refresh event for 1.14 firing, but it is a race so there could be other events that fire first
```
beliaev-maksim, line 62: > but it is a race so there could be other events that fire first

are they ?
upgrade is ensure to be fired the first

i-chvets:
LGTM.

i-chvets:
- Should we remove `istio-pilot:istio-pilot  istio-ingressgateway:istio-pilot` relation prior to removal of `istio-ingressgateway`?
- What should we watch for after deleting `istio-ingressgateway`? Should application be completely cleared from the list or `terminated` status is enough to start upgrades?

charms/istio-pilot/README.md
ca-scribner: According to [this discussion](https://chat.charmhub.io/charmhub/pl/wdatgrd557baxdec79wf9yjfpe) it would be a race

ca-scribner:
Removing the istio-pilot relation wont hurt anything but also shouldn't matter.  Removing the istio-gateway charm should implicitly remove the relation as well.

istio-gateway should be completely removed before upgrading to be safe

charms/istio-pilot/README.md
beliaev-maksim: but why do you need no retry here

if you refreshed to the right charm, then you expect to run upgrade hook. Retry means fire the same event, not the same code

DnPlas (commented):
Just a minor thing, thanks @ca-scribner

charms/istio-pilot/README.md
ca-scribner: the `--no-retry` here is to ignore the *previous* refresh.  So if you do:
* `juju refresh --channel something-broken`  --> Charm goes to error
* `juju refresh --channel something-working` --> Charm still in error from `something-broken`
* `juju resolved` --> resolves the error from `something-broken`, but Juju will try and rerun that same event again and go into error for the same reason
* `juju resolved --no-retry` --> resolves the error from `something-broken` and does not rerun that event, moving onto whatever is next (hopefully the refresh to `something-working`, but it could be some other event)

beliaev-maksim: @ca-scribner 
that is my point. Juju will resolve the error by firing the same event. But charm code is changed, so, next time it call the code from another charm

you have two charms that both can handle `upgrade` event

charms/istio-pilot/README.md
```diff
@@ -27,16 +27,25 @@ This charm supports in-place upgrades of the Istio control plane by up to one mi
...
-When upgrading istio-pilot, you should also upgrade istio-gateway at the same time to the same version.  These charms are only tested against each other at the same version.
+### Typical Upgrade Procedure
 
-If you need to upgrade across multiple versions (say from Istio 1.11.0 to Istio 1.16.0), refresh the charm one minor version at a time.  For example, starting with a deployed istio-pilot 1.11, you can do:
+When upgrading istio-pilot, you should **always remove the istio-gateway charm before upgrading** (some versions of `istio-pilot` do not upgrade properly if istio-gateway is deployed).  If you are upgrading one minor version, the procedure is:
+* `juju remove-application istio-ingressgateway`
+* `juju refresh istio-pilot --channel <desired-version>/stable`
+* `juju deploy istio-gateway --channel <desired-version>/stable --trust --config kind=ingress istio-ingressgateway`
+
+Where the istio-gateway config should match whatever you were using before upgrading.
```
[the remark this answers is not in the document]
DnPlas: Nothing to do, this section is already there.

DnPlas (approved):
LGTM, but let's resolve other conversations.

Issue #204, opened 2023-03-27
istio-pilot v1.12, v1.13, and v1.14 deploy incorrect ValidatingWebhookConfiguration during upgrade


ca-scribner:
`istioctl` v1.12, 1.13, and 1.14 have a bug where `istioctl upgrade` incorrectly creates a `ValidatingWebhookConfiguration` called `istiod-default-validator` that looks for `istiod` (the webhook's workload) in the default `istio-system` namespace, not the namespace we deploy `istio-pilot` to set (typically `kubeflow`). This results in a failing webhook immediately after upgrading to any of these versions and 404 errors raised when handling the `config-changed` event. 

This does not affect upgrading to istio v1.15+.  To fix this, we could patch the misconfigured webhook configuration after upgrade

PR #210, opened 2023-03-28
fix: patches misconfigured ValidatingWebhook on Install

ca-scribner:
This extends the patch for #204 to fix install events.  Prior to this patch, istioctl for v1.12-1.14 creates a ValidatingWebhookConfiguration that points to istiod in the istio-system namespace rather than the namespace where istiod is deployed.

### DIFF
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -105,6 +105,7 @@ class Operator(CharmBase):
 
     def install(self, event):
         """Install charm."""
+        self._log_and_set_status(MaintenanceStatus("Deploying Istio control plane"))
 
         subprocess.check_call(
             [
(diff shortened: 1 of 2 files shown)


charms/istio-pilot/src/charm.py
```diff
@@ -118,6 +119,13 @@ def install(self, event):
             ]
         )
 
+        # Patch any known issues with the install
+        # Istioctl v1.12-1.14 have a bug where the validating webhook is not deployed to the
+        # correct namespace (see https://github.com/canonical/istio-operators/issues/204)
+        # This has no effect if the webhook does not exist or is already correct
+        self._log_and_set_status(MaintenanceStatus("Patching webhooks"))
```
DnPlas, line 126: (:

DnPlas (approved):

DnPlas (approved):

PR #218, opened 2023-03-29
fix: clarify language in istio-gateway BlockedStatus

### LINKED ISSUE #217: istio-gateway's BlockStatus message when missing the istio-pilot relation is misleading
When istio-gateway is missing its required relation to istio-pilot, it goes to BlockedStatus with the message "Waiting for istio-pilot relation".  This feels a bit misleading as user intervention is required.  

https://github.com/canonical/istio-operators/blob/2ea88acb2abcde22c1682980c9cfbf6ab6e2b727/charms/istio-gateway/src/charm.py#L49

Suggested solution: change the wording to "Add required relation to istio-pilot" or similar.

ca-scribner:
Closes #217


charms/istio-gateway/src/charm.py
```diff
@@ -46,7 +46,7 @@ def start(self, event):
             return
 
         if not self.model.relations["istio-pilot"]:
-            self.model.unit.status = BlockedStatus("Waiting for istio-pilot relation")
+            self.model.unit.status = BlockedStatus("Add required relation to istio-pilot")
```
i-chvets, line 49: Looks weird to me as status message. It is more an instruction. May be "Waiting for missing istio-pilot relation" should be better?

beliaev-maksim: @i-chvets no
that is the point. Blocked status is a status which requires manual operation from human operator. Thus, we give an instruction what to do in order to unblock

beliaev-maksim (approved):

charms/istio-gateway/src/charm.py
ca-scribner: It shouldn't say "waiting" like it did before, but I'm kinda on the fence about these action statuses that you're suggesting @beliaev-maksim. they come across odd grammatically. maybe we can discuss in daily

i-chvets (approved):
Approved.

i-chvets (dismissed):
Approved.

beliaev-maksim (approved):
LGTM
please check that message fits into the window. If not trim `required`

PR #220, opened 2023-03-30
feat: make upgrade_charm handler wait on Istio's deployment during upgrades

### LINKED ISSUE #219: istio-pilot's upgrade_charm handler does not wait for upgrade to complete, resulting in race condition
When refreshing istio-pilot, the upgrade_charm event handler uses `istioctl` to do the upgrade of Istio.  Previously, the upgrade_charm handler would return as soon as `istioctl upgrade` completed, resulting in the charm going Active/Idle.  This results in a race condition because `istioctl upgrade` returns after it has **submitted the updated resources to kubernetes**, not after those updated resources have actually been rolled out.  This means that the charm goes to Active/Idle before the workload.  

This is particularly problematic when refreshing across a few versions in succession, because the successive upgrade_charm handlers will assert that istio-pilot can only upgrade across one minor version at a time.  For example:

0. (starting on istio-pilot 1.11)
1. `juju refresh istio-pilot --channel 1.12/stable`
2. `juju refresh istio-pilot --channel 1.13/stable`

Step (1) can go to Active/Idle before 1.12 is fully rolled out.  If we immediately execute step (2), the version checking logic of (2) will still see Istio v1.11 in the cluster and halt the upgrade process due to a version mismatch (cannot upgrade more than one minor version at a time).  This should resolve itself in time when Juju retries the handler, but still gives a confusing message to the user.

ca-scribner:
Closes #219.  Previously, upgrade_charm would initiate an Istio upgrade using `istioctl` and then proceed to Active/Idle regardless of whether that upgrade was complete.  Added here is a check in upgrade_charm to wait on this upgrade procedure until Istio registers as as updated to the new version.


charms/istio-pilot/README.md
```diff
@@ -72,3 +74,6 @@ If you accidentally attempt a downgrade, the istio-pilot charm will go to error
...
+[^2]: Note that in an older version of the istio-pilot upgrade handler, the charm would progress to Active/Idle as soon as the Kubernetes objects for Istio were updated, regardless of whether Kubernetes was still rolling out the new version of Istio's deployment.  This has been fixed in all istio-pilot charms hosted in CharmHub, but if you've pulled an older version could result in a race condition when refreshing across multiple charm versions quickly.  For example, if you refreshed from 1.12->1.13 and then quickly refreshed again from 1.13->1.14, the Istio deployments for 1.13 might still be rolling out and thus the istio-pilot charm for 1.14 might raise an error saying it cannot upgrade from 1.12 to 1.14.  If this occurs, use Kubernetes to check the rollout status for the `istiod` deployment and wait for it to complete.  Once that is complete, use `juju resolved istio-pilot` to rerun the upgrade handler for version 1.14 and continue your upgrade.
```
i-chvets, line 79: I would add here how to access those logs, i.e. `juju debug-log`.

charms/istio-pilot/README.md
```diff
@@ -72,3 +74,6 @@ If you accidentally attempt a downgrade, the istio-pilot charm will go to error
...
+[^2]: Note that in an older version of the istio-pilot upgrade handler, the charm would progress to Active/Idle as soon as the Kubernetes objects for Istio were updated, regardless of whether Kubernetes was still rolling out the new version of Istio's deployment.  This has been fixed in all istio-pilot charms hosted in CharmHub, but if you've pulled an older version it could result in a race condition when refreshing across multiple charm versions quickly.  For example, if you refreshed from 1.12->1.13 and then quickly refreshed again from 1.13->1.14, the Istio deployments for 1.13 might still be rolling out and thus the istio-pilot charm for 1.14 might raise an error saying it cannot upgrade from 1.12 to 1.14.  If this occurs, use Kubernetes to check the rollout status for the `istiod` deployment and wait for it to complete.  Once that is complete, use `juju resolved istio-pilot` to rerun the upgrade handler for version 1.14 and continue your upgrade.  Throughout this process, you can also check the `juju debug-log istio-pilot/0` to see detailed information about any errors during upgrade.
```
beliaev-maksim, line 79: if it is fixed for all the charms now, then maybe we do not need this note?
may cause confusions

charms/istio-pilot/src/charm.py
```diff
@@ -622,5 +636,34 @@ def _validate_upgrade_version(versions) -> bool:
     return True
 
 
+def _wait_for_update_rollout(
+    istioctl: Istioctl, retry_strategy: tenacity.Retrying, logger: logging.Logger
+):
+    for attempt in retry_strategy:
```
beliaev-maksim, line 642: how many times will it try until give up?

beliaev-maksim: just 5min?

charms/istio-pilot/src/charm.py
```diff
@@ -42,6 +43,13 @@
     "recommendations."
 )
 
+# Helper to retry calling a function over 300 seconds
+RETRY_FOR_15_MINUTES = tenacity.Retrying(
+    stop=tenacity.stop_after_delay(60 * 15),
```
beliaev-maksim, line 48: 60*5=300 
60*15=900

:D

ca-scribner: haha this is a typo in the comment.  Before it was 300, but bumped to 1500 just in case.  Fixed the comment

charms/istio-pilot/src/charm.py
ca-scribner: yeah 15 minutes

charms/istio-pilot/README.md
ca-scribner: Yeah as I wrote it, I felt this way too.  @i-chvets what do you think?  If someone updated between Wednesday and tomorrow when this ships they'll have the funny behaviour you see, but after that they'll always get the latest charm.  I wonder if we should just remove this

That said, this is a doc change that only needs to be in the main repo, so we dont need to hold up this functional change while we debate the doc change

charms/istio-pilot/src/charm.py
beliaev-maksim: from the code itself it is not that trivial to see. I mean "for attempt in retry", sounds like there will be several attempts. May be a short comment will be helpful

ca-scribner: every 2s for 15 minutes

i-chvets (approved):
Approved.

charms/istio-pilot/src/charm.py
beliaev-maksim: gotcha

charms/istio-pilot/README.md
i-chvets: Yes. We probably can get rid of it.

PR #222, opened 2023-03-30
backport: upgrade_charm handler fixes from #220 to `track/1.12`

### LINKED ISSUE #219: istio-pilot's upgrade_charm handler does not wait for upgrade to complete, resulting in race condition

ca-scribner:
Closes #219.  Previously, upgrade_charm would initiate an Istio upgrade using `istioctl` and then proceed to Active/Idle regardless of whether that upgrade was complete.  Added here is a check in upgrade_charm to wait on this upgrade procedure until Istio registers as as updated to the new version.


i-chvets (approved):
Approved.

PR #233, opened 2023-04-12
refactor istio-pilot event handling to use a central reconcile function

ca-scribner:
This is part of a larger effort to refactor istio-pilot's event handling, as described [in jira](https://warthogs.atlassian.net/browse/KF-728).  This merges into a feature branch and covers only part of the total effort. 

This PR:
* removes all code that is to be replaced during the istio-pilot event handling refactor
* removes all unit tests (these will be added back in in stages, but they require a signifcant refactor to match the event handling changes)
* refactors `Operator.__init__` to use `self.reconcile` for most event handling
* adds unit tests that ensure we subscribe to the events we expect in `Operator.__init__` (similar to how istio-pilot used to test this)
* adds tests that assert leadership
* adds nearly-functional `Operator.reconcile` for handling most charm events and the stubs of helper functions that reconcile requires to operate

### Notes to reviewers

Please focus on:
* `Operator.__init__`: specifically, does it look like the proper charm events are captured?
* `Operator.reconcile`: specifically, does the logic make sense?  Assuming the helpers do what they say, does this work?
* the new unit tests (note that these are refactors of previous tests)

This PR is not fully functional, but the tests added should work and cover what they say.


charms/istio-pilot/src/charm.py
```diff
@@ -136,6 +145,64 @@ def install(self, event):
...
+            interfaces = self._get_interfaces()
+            self._reconcile_ingress(interfaces['ingress'], event)
+        except ErrorWithStatus as err:
+            # One or more related applications resulted in an error
+            handled_errors.append(err)
+            self._log_and_set_status(err.status)
+
+        # TODO: If we have and handled_errors, report them
+        self._report_handled_errors(handled_errors)
+
```
i-chvets, line 203: Where are we going to raise the "worst" error?

ca-scribner: I'm thinking this handler does something like:
```python
def _report_handled_errors(self, handled_errors):
  worst_error = get_worst_error(handled_errors)
  worst_error_type = worst_error.class  # Or however you get this...
  headline_message = f"During operation we encountered {len(handled_errors)} errors.  See logs for details."
  self.unit.status = worst_error_type(f"During operation we encountered {len(handled_errors)} handled errors.  See logs for details.")
  self.log.info(f"During operation we encountered {len(handled_errors)} handled errors:")

  for error in handled_errors:
    self.log.info(f"Handled error of type {error.status}: {error.message}")
```

what do you think?

i-chvets: Look good. I think, we should use `log.error` though. Also, do we need to explicitly call `raise` for that "worst" error?

i-chvets (approved):
Approved.

PR #234, opened 2023-04-12
add unit tests for upgrade to refactored istio-pilot

ca-scribner:
This is based on #233.  Recommended that this be reviewed after #233 is merged as that accounts for most of the changes.  This is part of a larger effort to refactor istio-pilot's event handling, as described [in jira](https://warthogs.atlassian.net/browse/KF-728). This merges into a feature branch and covers only part of the total effort.

This PR re-adds the istio-pilot upgrade tests.  These are ported from the previous version of istio-pilot but rearranged to make the test suite clearer.  They also include small modifications to match the new charm structure, but are generally the same.


charms/istio-pilot/src/charm.py
```diff
@@ -434,8 +446,9 @@ def _patch_istio_validating_webhook(self):
             "Attempting to patch istiod-default-validator webhook to ensure it points to"
             " correct namespace."
         )
+        lightkube_client = Client()
         try:
-            vwc = self.lightkube_client.get(
+            vwc = lightkube_client.get(
```
i-chvets, line 451: what is `vwc`?

i-chvets (approved):
Approved.

PR #235, opened 2023-04-12
add istio-gateway helpers to istio-pilot event handler rewrite

ca-scribner:
This is based on #234.  Recommended that this be reviewed after #234 is merged as that accounts for most of the changes.  This is part of a larger effort to refactor istio-pilot's event handling, as described [in jira](https://warthogs.atlassian.net/browse/KF-728). This merges into a feature branch and covers only part of the total effort.

Includes:
* helpers that check on istio-gateway's service
* tests for helpers
* added a config option for the istio-gateway's service name to make things easier to test and deploy

### DIFF
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -358,19 +358,6 @@ class Operator(CharmBase):
         )
         return svc
 
-    def _is_gateway_service_up(self):
-        """Returns True if the ingress gateway service is up, else False."""
-        # TODO: This should really be something provided via a relation to istio-gateway, where it
-        #  tells us if things are working.
-        svc = self._get_gateway_service()
-
-        if svc.spec.type == "NodePort":
-            # TODO: do we need to interrogate this further for status?
-            return True
-        if _get_gateway_address_from_svc(svc) is not None:
-            return True
-        return False
-
     def _send_gateway_info(self):
         """Sends gateway information to all related apps."""
         # TODO: Can any of this be put into the lib?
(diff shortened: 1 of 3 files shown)


charms/istio-pilot/src/charm.py
```diff
@@ -347,6 +346,31 @@ def _get_ingress_auth_data(self) -> dict:
...
+        #  from the istio-gateway.  Providing here as a config at least makes this less rigid than
+        #  assuming the name.
+        # TODO: What happens if this service does not exist?  We should check on that and then add
+        #  tests to confirm this works
+        svc = self.lightkube_client.get(
+            Service, name=self.model.config["gateway-service-name"], namespace=self.model.name
+        )
+        return svc
+
+    def _is_gateway_service_up(self):
```
beliaev-maksim, line 361: maybe convert to property ?

ca-scribner: I'm never sure when to use a property.  Do you have a guideline to recommend?  

My rule of thumb is to do it when there's essentially no computation behind the item.  This one has very trivial computation behind it, but lots of logic across a few helpers.  So I'm torn :D

i-chvets:
LGTM. Let me know when approval is needed.

i-chvets (approved):
Approved.

charms/istio-pilot/src/charm.py
beliaev-maksim: I think the whole idea of properties to have computation behind :)
otherwise you use just class attribute

ca-scribner: done!

beliaev-maksim (approved):

PR #237, opened 2023-04-18
refactor: include provider class in gateway-info library

DnPlas:
This library can be used by other charms that can provide gateway information. The previous implementation of the library did not include the provider logic. This commit extends the API to be able to instantiate the provider class in other charms other than istio-pilot.

TODO:
- [x] Add unit tests

### DIFF
--- a/charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
+++ b/charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
@@ -156,6 +156,7 @@ class GatewayRequirer(Object):
         Raises:
             GatewayRelationDataMissingError: if data is missing or incomplete
             GatewayRelationMissingError: if there is no related application
+            ops.model.TooManyRelatedAppsError: if there is more than one related application
         """
         # Raise if there is no related applicaton
         if not relation:
--- a/charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
+++ b/charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
@@ -78,7 +78,7 @@ class ProviderCharm(self):
 The data shared by this library has only two items:
 
 * gateway_name: the name of the Gateway the provider knows about. It corresponds to
-the `name` field in the Gateway definition:
+the `name` field in the Gateway definition. In the following example `gateway_name` value will be equal to `my-gateway`
 
 ```yaml
 apiVersion: networking.istio.io/v1alpha3
--- a/charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
+++ b/charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
@@ -54,7 +54,6 @@ Class RequirerCharm(CharmBase):
 provides:
   gateway-info:
     interface: istio-gateway-info
-    limit: 1
 ```
 
 ### Instantiate the GatewayProvider class in charm.py
@@ -133,7 +132,8 @@ class GatewayRequirer(Object):
         self.requirer_charm = requirer_charm
         self.relation_name = relation_name
 
-    def _relation_preflight_checks(self, relation: Relation) -> None:
+    @staticmethod
+    def _relation_preflight_checks(relation: Relation) -> None:
         """Series of checks for the relation and relation data.
         Args:
             relation (Relation): the relation object to run the checks on
@@ -206,16 +206,17 @@ class GatewayProvider(Object):
             gateway_namespace(str): the namespace of the Gateway the provider knows about
         """
         # Update the relation data bag with localgateway information
-        relation = self.model.get_relation(self.relation_name)
+        relations = self.model.relations[self.relation_name]
 
         # Raise if there is no related applicaton
-        if not relation:
+        if not relations:
             raise GatewayRelationMissingError()
 
         # Update relation data
-        relation.data[self.provider_charm.app].update(
-            {
-                "gateway_name": gateway_name,
-                "gateway_namespace": gateway_namespace,
-            }
-        )
+        for relation in relations:
+            relation.data[self.provider_charm.app].update(
+                {
+                    "gateway_name": gateway_name,
+                    "gateway_namespace": gateway_namespace,
+                }
+            )
--- a/charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
+++ b/charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
@@ -186,6 +186,7 @@ class GatewayRequirer(Object):
             GatewayRelationMissingError: if there is no related application
         """
         # Run pre-flight checks
+        # Raises TooManyRelatedAppsError if related to more than one app
         relation = self.model.get_relation(self.relation_name)
         self._relation_preflight_checks(relation=relation)
 
--- a/charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
+++ b/charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
@@ -72,6 +72,22 @@ class ProviderCharm(self):
         except GatewayRelationError as error:
             raise <your preferred exception with a message> from error
 ```
+
+## Relation data
+
+The data shared by this library has only two items:
+
+* gateway_name: the name of the Gateway the provider knows about. It corresponds to
+the `name` field in the Gateway definition:
+
+```yaml
+apiVersion: networking.istio.io/v1alpha3
+kind: Gateway
+metadata:
+  name: my-gateway
+```
+
+* gateway_namespace: the namespace where the Gateway is deployed.
 """
 
 import logging
--- a/charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
+++ b/charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
@@ -1,46 +1,84 @@
-"""Library for sharing istio gateway information
+#!/usr/bin/env python3
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
 
-This library wraps the relation endpoints using the `istio-gateway-name`
-interface. It provides a Python API for both requesting and providing 
-gateway information.
+"""Library for sharing Istio Gateway(s) information
 
+This library offers a Python API for providing and requesting information about
+Istio Gateway(s) by wrapping the `gateway-info` relation endpoints.
+The default relation name is `gateway-info` and it's recommended to use that name,
+though if changed, you must ensure to pass the correct name when instantiating the
+provider and requirer classes, as well as in metadata.yaml.
+ 
 ## Getting Started
 
-### Fetch library with charmcraft
-You can fetch the library using the following commands with charmcraft.
+### Fetching the library with charmcraft
+
+Using charmcraft you can:
 ```shell
-cd some-charm
 charmcraft fetch-lib charms.istio_pilot.v0.istio_gateway_info
 ```
+
+## Using the library as requirer
+
 ### Add relation to metadata.yaml
 ```yaml
 requires:
-    gateway-info:
-        interface: istio_gateway_info
-        limit: 1
+  gateway-info:
+    interface: istio-gateway-info
+    limit: 1
 ```
 
-### Initialise the library in charm.py
-```python
-from charms.istio_pilot.v0.istio_gateway_info import GatewayProvider, GatewayRelationError
+### Instantiate the GatewayRequirer class in charm.py
 
-Class SomeCharm(CharmBase):
+```python
+from charms.istio_pilot.v0.istio_gateway_info import GatewayRequirer, GatewayRelationError
+
+Class RequirerCharm(CharmBase):
     def __init__(self, *args):
-        self.gateway = GatewayProvider(self)
+        self.gateway = GatewayRequirer(self)
         self.framework.observe(self.on.some_event_emitted, self.some_event_function)
 
     def some_event_function():
         # use the getter function wherever the info is needed
         try:
             gateway_data = self.gateway_relation.get_relation_data()
-            except GatewayRelationError as error:
-            ...
+        except GatewayRelationError as error:
+            raise <your preferred exception> from error
+```
+
+## Using the library as provider
+
+### Add relation to metadata.yaml
+```yaml
+provides:
+  gateway-info:
+    interface: istio-gateway-info
+    limit: 1
+```
+
+### Instantiate the GatewayProvider class in charm.py
+
+```python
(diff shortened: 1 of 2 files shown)


beliaev-maksim:
@DnPlas 
any tests for the lib?

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
```diff
@@ -50,15 +88,41 @@ def some_event_function():
...
+        relation_name (str): the name of the relation
+        requirer (bool, optional): set to True if the check is run in the requirer class
+    Raises:
+        GatewayRelationMissingError when there is no relation between two units
+        GatewayRelationDataMissingError when there is no data in the data bag
+    """
+    # Do nothing if the unit is not leader
+    if not model.unit.is_leader():
+        logger.info("This unit is not leader, no action will be taken.")
+        raise RuntimeError("Unit is not leader")
```
ca-scribner, line 116: This feels like something the charm should be responsible for, not the library.  The charm would know if leadership matters for whether it uses this relation's data, we shouldn't impose that on the charm.

For example, if a user of this check offers HA, this means the non-leader units cannot check if they have relation data

ca-scribner: ah, I see the old one applied this check as well. I'm on the fence on whether we should immediately remove or treat it as something new for later

beliaev-maksim: overall check for leader is fine. The question if we should raise or not
https://github.com/canonical/prometheus-k8s-operator/blob/9d386da8023f91049252adab5e1db5fb9845251e/lib/charms/prometheus_k8s/v0/prometheus_scrape.py#L1870

beliaev-maksim (commented):
draft pre-approved

wanna unskip tests :)

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
```diff
@@ -82,44 +146,89 @@ def __init__(self, message):
...
+        super().__init__(requirer_charm, relation_name)
+        # TODO: remove this attribute, we are not really using it at all.
+        self.requirer_charm = requirer_charm
+        self.relation_name = relation_name
 
+    def get_relation_data(self) -> dict:
+        """Returns a dictionary with the Gateway information.
+
+        Raises:
+            GatewayRelationDataMissingError: if data is missing attributes
```
ca-scribner, line 170: Since this calls `relation_preflight_checks`, it might raise anything from there.  Maybe say this?


```suggestion
        Raises:
            GatewayRelationDataMissingError: if data is missing
            GatewayRelationMissingError: if there is no related application
```

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
```diff
@@ -50,15 +88,41 @@ def some_event_function():
...
+    model: Model, relation_name: str, requirer: bool = False
+) -> None:
+    """Series of checks for ensuring the relation is properly established.
+    Args:
+        model (Model): the juju model as seen by the provider/requirer unit
+        relation_name (str): the name of the relation
+        requirer (bool, optional): set to True if the check is run in the requirer class
+    Raises:
+        GatewayRelationMissingError when there is no relation between two units
+        GatewayRelationDataMissingError when there is no data in the data bag
```
ca-scribner, line 111: I think RelationMissing cares about related applications, not units.  Also changing this to be consistent with `get_relation_data()`'s docstring
```suggestion
        GatewayRelationDataMissingError: if data is missing
        GatewayRelationMissingError: if there is no related application
```

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
```diff
@@ -50,15 +88,41 @@ def some_event_function():
...
-INTERFACE_NAME = "istio-gateway-info"
+# Default relation and interface names. If changed, consistency must be kept
+# across the provider and requirer.
+DEFAULT_RELATION_NAME = "gateway-info"
+DEFAULT_INTERFACE_NAME = "istio-gateway-info"
 
 logger = logging.getLogger(__name__)
 
 
+def relation_preflight_checks(
```
beliaev-maksim, line 101: to the point of Andrew https://github.com/canonical/istio-operators/pull/237/files#r1170064533

maybe make this method private ?

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
ca-scribner: hold up - I think I'm wrong about the data being in units vs applications.  Maybe that language should stay as referring to a unit

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
```diff
@@ -82,44 +146,89 @@ def __init__(self, message):
...
+
+        # This charm should only establish a relation with exactly one unit
+        # the following extracts exactly one unit from the set that's
+        # returned by relation.data
+        relation = self.model.get_relation(self.relation_name)
+        remote_units = relation.units
+        provider_unit = list(remote_units)[0]
+
+        # Get relation data
+        relation_data = relation.data[provider_unit]
```
ca-scribner, line 183: I think here we are reading data from a unit of the related application, whereas previously the relation read data from the application part of the data bag.  This wont be backward compatible with the previous lib

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
ca-scribner: ok I think the new version of the lib is reading from unit, but the old version is reading from app.  See [this conversation](https://github.com/canonical/istio-operators/pull/237/files#r1170073824)

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
```diff
@@ -50,15 +88,41 @@ def some_event_function():
...
+    if not model.unit.is_leader():
+        logger.info("This unit is not leader, no action will be taken.")
+        raise RuntimeError("Unit is not leader")
+
+    relation = model.get_relation(relation_name)
+    if not relation:
+        raise GatewayRelationMissingError()
+
+    if requirer and not relation.data:
+        raise GatewayRelationDataMissingError("No data found in relation data bag.")
```
ca-scribner, line 123: should this assert that data exists, or that *valid* data exists (eg check for the expected data keys, like done [here](https://github.com/canonical/istio-operators/pull/237/files#diff-85507bb21a0ce73a75d60021f5ae8970194bd2e968a5e1fc0b954a41acae3adeR186-R193).  I think those checks could be pulled into here

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
DnPlas: Will change from unit to app.

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
DnPlas: +1 on private method, not sure what you mean about the first phrase.

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
DnPlas: I think I resolved this one by removing the check. Please review the updated version of the code.

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
DnPlas: Updated

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
ca-scribner, line 1: Is the data passed through this relation documented somewhere here?  I didn't see it in the docstring at the top, but I might have missed it.  

If not, let's have a clear description of what fields are available and what they mean in the docstring

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
```diff
@@ -1,46 +1,84 @@
...
+```
+
+## Using the library as provider
+
+### Add relation to metadata.yaml
+```yaml
+provides:
+  gateway-info:
+    interface: istio-gateway-info
+    limit: 1
```
beliaev-maksim, line 57: is provider always limited to one ?

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
```diff
@@ -82,44 +115,107 @@ def __init__(self, message):
...
+        if not relation:
+            raise GatewayRelationMissingError()
+
+        # Update relation data
+        relation.data[self.provider_charm.app].update(
+            {
+                "gateway_name": gateway_name,
+                "gateway_namespace": gateway_namespace,
+            }
+        )
```
ca-scribner, line 221: Is there a dimension missing here?  I think `relation` will be a list of data bags, one for each related application, wont it?  For example, this application is providing the gateway-info relation to two different related applications, how is that handled here?

ca-scribner: We see this in the old relation provider [here](https://github.com/canonical/istio-operators/blob/b2f820fff6d24d692d4135d041867e3cbff30efc/charms/istio-pilot/src/istio_gateway_info_provider.py#L15) - they loop through the relations.  

I think the same is valid on the Provider side as well, just from looking at the old Provider code [here](https://github.com/canonical/istio-operators/pull/237/files#diff-85507bb21a0ce73a75d60021f5ae8970194bd2e968a5e1fc0b954a41acae3adeL101) which had an index

charms/istio-pilot/tests/unit/test_gateway_info.py
```diff
@@ -47,3 +67,122 @@ def test_get_relation_data_passes(requirer_charm_harness, mocker):
...
+    # Instantiate GatewayRequirer class
+    requirer_charm_harness.charm.gateway_requirer = GatewayRequirer(
+        requirer_charm_harness.charm, relation_name=TEST_RELATION_NAME
+    )
+
+    requirer_charm_harness.add_relation(TEST_RELATION_NAME, "app")
+    requirer_charm_harness.add_relation(TEST_RELATION_NAME, "app2")
+
+    with pytest.raises(TooManyRelatedAppsError):
+        requirer_charm_harness.charm.gateway_requirer.get_relation_data()
```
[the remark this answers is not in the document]
DnPlas: This exception will be raised automatically when you join more than two apps. From the docs: "Raised by [Model.get_relation()](https://ops.readthedocs.io/en/latest/index.html?highlight=TooManyRelatedAppsError#ops.model.Model.get_relation) if there is more than one related application."

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
DnPlas: Yes, you are right. Updated.

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
```diff
@@ -72,6 +72,22 @@ def _some_event_handler(self, ...):
         except GatewayRelationError as error:
             raise <your preferred exception with a message> from error
 ```
+
+## Relation data
+
+The data shared by this library has only two items:
+
+* gateway_name: the name of the Gateway the provider knows about. It corresponds to
+the `name` field in the Gateway definition:
```
beliaev-maksim, line 81:
```suggestion
the `name` field in the Gateway definition. In the following example `gateway_name` value will be equal to `my-gateway`
```

beliaev-maksim (approved):

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
```diff
@@ -82,44 +130,109 @@ def __init__(self, message):
...
-        gateway = self.model.relations[self.relation_name]
-        if len(gateway) == 0:
+    @staticmethod
+    def _relation_preflight_checks(relation: Relation) -> None:
+        """Series of checks for the relation and relation data.
+        Args:
+            relation (Relation): the relation object to run the checks on
+        Raises:
+            GatewayRelationDataMissingError: if data is missing or incomplete
+            GatewayRelationMissingError: if there is no related application
```
ca-scribner, line 158:
```suggestion
            GatewayRelationDataMissingError: if data is missing or incomplete
            GatewayRelationMissingError: if there is no related application
            ops.model.TooManyRelatedAppsError: if there is more than one related application
```

beliaev-maksim: @ca-scribner good question if we should tell about raises from other lib.

But from other side, since we applied the method and allow to pass the error through I am +1

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
```diff
@@ -82,44 +130,109 @@ def __init__(self, message):
...
+
+    def get_relation_data(self) -> dict:
+        """Returns a dictionary with the Gateway information.
+
+        Raises:
+            GatewayRelationDataMissingError: if data is missing entirely or some attributes
+            GatewayRelationMissingError: if there is no related application
+        """
+        # Run pre-flight checks
+        relation = self.model.get_relation(self.relation_name)
```
ca-scribner, line 189: We count on `get_relation` to validate that we have <=1 related application.  That feels like something that could be easily missed - maybe we can add a comment here?
```suggestion
        # Raises TooManyRelatedAppsError if related to more than one app
        relation = self.model.get_relation(self.relation_name)
```

charms/istio-pilot/tests/unit/test_gateway_info.py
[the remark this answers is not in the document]
ca-scribner: k thanks!

ca-scribner (approved):
lgtm!

charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py
ca-scribner: yeah I think where it is enforcing a key function (that this requirer shouldn't have more than 1 app) we should probably say so here

beliaev-maksim (approved):

ca-scribner (approved):
lgtm...again!

PR #239, opened 2023-04-18
Update README.md

beliaev-maksim:
we do not support charm store anymore. Thus, should not deploy bundle from there. Also, bundle is out of date


DnPlas (approved):

DnPlas (commented):
Please add a follow up PR for adding right instructions.

PR #240, opened 2023-04-19
refactor: use refactored gateway-info in handle_gateway_info event handler

DnPlas:
Use the latest version of gateway-info library. 

This PR should be merged after #237

### DIFF
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -8,7 +8,7 @@ import yaml
 from charmed_kubeflow_chisme.exceptions import GenericCharmRuntimeError
 from charms.grafana_k8s.v0.grafana_dashboard import GrafanaDashboardProvider
 from charms.istio_pilot.v0.istio_gateway_info import (
-    RELATION_NAME,
+    DEFAULT_RELATION_NAME,
     GatewayProvider,
     GatewayRelationMissingError,
 )
@@ -106,7 +106,7 @@ class Operator(CharmBase):
         # FIXME: Calling handle_gateway_relation on update_status ensures gateway information is
         # sent eventually to the related units, this is temporal and we should find a way to
         # ensure all event handlers are called when they are supposed to.
-        for event in [self.on[RELATION_NAME].relation_changed, self.on.update_status]:
+        for event in [self.on[DEFAULT_RELATION_NAME].relation_changed, self.on.update_status]:
             self.framework.observe(event, self.handle_gateway_info_relation)
         self.framework.observe(self.on["istio-pilot"].relation_changed, self.send_info)
         self.framework.observe(self.on["ingress"].relation_changed, self.handle_ingress)
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -7,6 +7,11 @@ import tenacity
 import yaml
 from charmed_kubeflow_chisme.exceptions import GenericCharmRuntimeError
 from charms.grafana_k8s.v0.grafana_dashboard import GrafanaDashboardProvider
+from charms.istio_pilot.v0.istio_gateway_info import (
+    RELATION_NAME,
+    GatewayProvider,
+    GatewayRelationMissingError,
+)
 from charms.prometheus_k8s.v0.prometheus_scrape import MetricsEndpointProvider
 from jinja2 import Environment, FileSystemLoader
 from lightkube import Client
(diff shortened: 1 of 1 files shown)


charms/istio-pilot/src/charm.py
```diff
@@ -293,23 +293,38 @@ def handle_default_gateway(self, event):
...
-        if is_gateway_created:
-            self.gateway.send_gateway_relation_data(
-                self.app, self.model.config["default-gateway"], self.model.name
-            )
-        else:
+
+        # Raise if the default-gateway is not created, it must be present always
+        if not is_gateway_created:
             self.log.info("Gateway is not created yet. Skip sending gateway relation data.")
+            raise GenericCharmRuntimeError("default-gateway not present in cluster.")
```
beliaev-maksim, line 315: I do not think we need to raise the error here. That will put the unit into error state

beliaev-maksim (changes requested):

charms/istio-pilot/src/charm.py
DnPlas: I was thinking this a lot, but to me if there is no default-gateway, istio-pilot is not doing its job correctly. I was not sure to raise this exception here because this should be handled elsewhere in the charm code, but as a check is better.

ca-scribner: I think this gets refactored by the stuff I'm doing right now anyway, so not important to battle it out here imo.  We can debate it in a day on mine if needed :)

charms/istio-pilot/src/charm.py
```diff
@@ -293,23 +297,38 @@ def handle_default_gateway(self, event):
...
+        try:
+            self.gateway_provider.send_gateway_relation_data(
+                gateway_name=self.model.config["default-gateway"],
+                gateway_namespace=self.model.name,
+            )
+        except GatewayRelationMissingError:
+            # If there is no relation, we cannot send data. In reality, the execution
+            # should not get this far.
+            self.unit.status = BlockedStatus("Please add gateway-info relation.")
+            return
```
ca-scribner, line 331: I don't think we need this here.  This charm provides gateway-info information to anyone who is listening.  If nobody is listening, nothing is blocked it just means a user has chosen not to use this function.

ca-scribner (changes requested):

charms/istio-pilot/src/charm.py
beliaev-maksim: need to understand whose code will land faster. I also do not want to block Daniela

charms/istio-pilot/src/charm.py
DnPlas: But if you are observing the relation event and handling it, you should expect the relation (see also my comment one line above).

charms/istio-pilot/src/charm.py
DnPlas: Well, this change won't block any kserve work. I only needed the library, this PR is a consequence of that change and only applies to istio-pilot. We can have minimal changes in this PR and let @ca-scribner do the rest.

charms/istio-pilot/src/charm.py
ca-scribner: You should expect the relation, but not expect that applications are related to it.  For example, doing:

```
juju deploy istio-pilot
juju deploy istio-gateway
juju relate istio-pilot istio-gateway
```
should result in istio-pilot and istio-gateway being green right?  But with this check, we'll actually go to BlockedStatus until we do:
```
juju deploy something-that-needs-gateway-info
juju relate istio-pilot:gateway-info something-that-needs-gateway-info:gateway-info
```

which I don't think is what we want.  

The problem is really stemming from [the Provider class](https://github.com/canonical/istio-operators/blob/b834927084bf26759813f9e2b86b40b23c05471b/charms/istio-pilot/lib/charms/istio_pilot/v0/istio_gateway_info.py#L230) which should not raise if there are no related applications, but instead should silently proceed, but we missed it in the other PR.

ca-scribner: You're right that I'll be touching this too in the next day or so so maybe its better for me to adjust it then, but with `GatewayRelationMissingError`s being raised in this situation the relation is not backward compatible (because istio-pilot will go to blocked in situations it used to not go to blocked for) and I'd argue semi-broken since basic deployments will go blocked.

charms/istio-pilot/src/charm.py
ca-scribner: yeah I say we resolve this thread and it'll get sorted out by end of week anyway

charms/istio-pilot/src/charm.py
ca-scribner: I might be misunderstanding this too, so if I am tell me I'm wrong :)  I'm a little confused why this hasn't broken the integration tests.  Although if I look at their [logs](https://github.com/canonical/istio-operators/actions/runs/4747082970/jobs/8431528979?pr=240), I'm not sure if they're passing or False negativing-ing

DnPlas: The only case where the exception is raised is if you are using `send_gateway_relation_data` without having the relation, which is less likely because this event handler is reacting to the `gateway_info_relation_*` event.

In the best interest of time, I have changed this PR with minimal changes, you can decide how to proceed later.

beliaev-maksim (approved):

PR #243, opened 2023-04-19
feat: add Gateway reconciliation/removal features

ca-scribner:
Adds:
* _reconcile_gateway
* _remove_gateway
* helpers to support these
* unit tests to test everything added

Note that this temporarily uses a version of charmed-kubeflow-chisme pulled from a specific commit in github.  This is required to enable KubernetesResourceHandler.reconcile() and .delete() features.  This will be removed when the updated KRH is approved and merged.

### DIFF
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -6,12 +6,15 @@ from typing import Optional
 
 import tenacity
 from charmed_kubeflow_chisme.exceptions import ErrorWithStatus, GenericCharmRuntimeError
-from charmed_kubeflow_chisme.kubernetes import KubernetesResourceHandler
+from charmed_kubeflow_chisme.kubernetes import (
+    KubernetesResourceHandler,
+    create_charm_default_labels,
+)
 from lightkube import Client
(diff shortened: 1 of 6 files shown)


charms/istio-pilot/src/charm.py
```diff
@@ -582,6 +625,18 @@ def _patch_istio_validating_webhook(self):
         )
         self.log.info("istiod-default-validator webhook successfully patched")
 
+    def _use_https(self):
+        if _xor(self.model.config["ssl-crt"], self.model.config["ssl-key"]):
+            # Fail if ssl is only partly configured as this is probably a mistake
+            raise ErrorWithStatus(
+                "Charm config for ssl-crt and ssl-key must either both be set or unset",
```
beliaev-maksim, line 632: sounds strange
xor does  either `(True & False) or (False & True)`

in message we say that it should be (True&True) or (False&False)

beliaev-maksim (approved):
first impression is good

but want Ivan's feedback as well

i-chvets (approved):
Approved.

PR #247, opened 2023-04-20
feat: add ingress relation handler to istio-pilot event handling refactor

ca-scribner:
This is based on #243. Recommended that this be reviewed after #243 is merged as that accounts for some of the deltas. This is part of a larger effort to refactor istio-pilot's event handling, as described [in jira](https://warthogs.atlassian.net/browse/KF-728). This merges into a feature branch and covers only part of the total effort.

* adds _get_ingress_data() to pulls route data from ingress relations
* adds _reconcile_ingress() to update, modify, and delete ingresses based on the related applications
* refactors the virtual_service.yaml.j2
* refactors gateway name/namespace handling into properties to make it shared between gateway and ingress handling
* adds unit tests for above new functions

### DIFF
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -15,7 +15,7 @@ from lightkube.core.exceptions import ApiError
 from lightkube.generic_resource import create_namespaced_resource
 from lightkube.resources.admissionregistration_v1 import ValidatingWebhookConfiguration
 from lightkube.resources.core_v1 import Secret, Service
-from ops.charm import CharmBase
+from ops.charm import CharmBase, RelationBrokenEvent
 from ops.main import main
 from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingStatus
 from packaging.version import Version
@@ -44,6 +44,7 @@ KRH_GATEWAY_SCOPE = "gateway"
 METRICS_PORT = 15014
 INGRESS_AUTH_RELATION_NAME = "ingress-auth"
 INGRESS_AUTH_TEMPLATE_FILES = ["src/manifests/auth_filter.yaml.j2"]
+INGRESS_RELATION_NAME = "ingress"
 ISTIOCTL_PATH = "./istioctl"
 ISTIOCTL_DEPOYMENT_PROFILE = "minimal"
 UPGRADE_FAILED_MSG = (
@@ -211,6 +212,7 @@ class Operator(CharmBase):
                 # TODO: Log here?
                 self._remove_gateway()
         except ErrorWithStatus as err:
+            # TODO: Is there anything to catch here?
             handled_errors.append(err)
 
         try:
@@ -382,6 +384,7 @@ class Operator(CharmBase):
             raise ErrorWithStatus(err, BlockedStatus)
 
         # Filter out data we sent out.
+        # TODO: Is this needed here?
         if ingress_auth_interface:
             ingress_auth_data = {
                 (rel, app): route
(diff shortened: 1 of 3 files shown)


charms/istio-pilot/src/charm.py
```diff
@@ -401,6 +412,39 @@ def _get_ingress_auth_data(self) -> dict:
...
+    def _get_ingress_data(self, event) -> dict:
+        """Retrieve the ingress relation data without touching other interface data.
+
+        This is a workaround to ensure that errors in other relation data, such as an incomplete
+        ingress-auth relation, do not block us from retrieving the ingress data.
+        """
+        try:
+            ingress_interface = get_interface(self, INGRESS_RELATION_NAME)
+        except NoVersionsListed as err:
+            raise ErrorWithStatus(err, WaitingStatus)
```
beliaev-maksim, line 424: should not this be `str(err)` ?

charms/istio-pilot/src/charm.py
```diff
@@ -401,6 +412,39 @@ def _get_ingress_auth_data(self) -> dict:
...
+
+        # Filter out data we sent out.
+        if ingress_interface:
+            routes = {
+                (rel, app): route
+                for (rel, app), route in sorted(
+                    ingress_interface.get_data().items(), key=lambda tup: tup[0][0].id
+                )
+                if app != self.app
+            }
```
beliaev-maksim, line 436: can we expand it to standard loops in favor of readability ?

charms/istio-pilot/src/charm.py
```diff
@@ -401,6 +412,39 @@ def _get_ingress_auth_data(self) -> dict:
...
+        else:
+            # If there is no ingress-auth relation, we have no data here
+            routes = {}
+
+        if isinstance(event, RelationBrokenEvent):
+            # The app-level data is still visible on a broken relation, but we
+            # shouldn't be keeping the VirtualService for that related app.
+            del routes[(event.relation, event.app)]
```
beliaev-maksim, line 444: maybe `pop` the item?
have some cultural issues with del :D

charms/istio-pilot/src/charm.py
ca-scribner: For convenience, ErrorWithStatus's `msg` arg is [always converted to a string](https://github.com/canonical/charmed-kubeflow-chisme/blob/592c8d8eba7d655f576c8ac8971451ffb31a4251/src/charmed_kubeflow_chisme/exceptions/_with_status.py#L16) in the constructor, so I think this is ok?

charms/istio-pilot/src/charm.py
ca-scribner: done!

charms/istio-pilot/src/charm.py
ca-scribner: done!

charms/istio-pilot/src/charm.py
```diff
@@ -773,6 +842,26 @@ def _wait_for_update_rollout(
...
+    routes = {}
+
+    if ingress_interface:
+        sorted_interface_data = sorted(
+            ingress_interface.get_data().items(), key=lambda tup: tup[0][0].id
+        )
+
+        for (rel, app), route in sorted_interface_data:
+            if app != this_app:
+                routes[(rel, app)] = route
```
beliaev-maksim, line 857: that block basically means that we pick the latest route and override in some cases ?

what is the case when it will happen ?

ca-scribner: The ingress_interface.get_data() returns a dict of format:
`(relation, application): route_data`
where application might be either us, or the other related application

We go through that data and pull all the data that is from other applications, ignoring the data that is from us.

I think it is conceptually similar to something like:
```python
for relation in all_relations:
  try:
    routes.pop(relation, self.app)
  except KeyError:
    # ignore
```

beliaev-maksim (approved):

PR #248, opened 2023-04-20
feat: add handler for istio-pilot's istio-pilot relation

ca-scribner:
This is based on #247. Recommended that this be reviewed after #247 is merged as that accounts for some of the deltas. This is part of a larger effort to refactor istio-pilot's event handling, as described [in jira](https://warthogs.atlassian.net/browse/KF-728). This merges into a feature branch and covers only part of the total effort.

Adds handler for istio-pilot's istio-pilot relation, plus tests.

### DIFF
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -51,6 +51,7 @@ METRICS_PORT = 15014
 INGRESS_AUTH_RELATION_NAME = "ingress-auth"
 INGRESS_AUTH_TEMPLATE_FILES = ["src/manifests/auth_filter.yaml.j2"]
 INGRESS_RELATION_NAME = "ingress"
+ISTIO_PILOT_RELATION_NAME = "istio-pilot"
 KRH_INGRESS_SCOPE = "ingress"
 VIRTUAL_SERVICE_TEMPLATE_FILES = ["src/manifests/virtual_service.yaml.j2"]
 ISTIOCTL_PATH = "./istioctl"
@@ -115,10 +116,8 @@ class Operator(CharmBase):
         # * relation_joined: because we send data to the other side whenever we see a related app
         # * relation_changed: because of SDI's data versioning model, which first agrees on the
         #                     schema version and then sends the rest of the data
-        # TODO: * relation_broken: is this needed?
         self.framework.observe(self.on["istio-pilot"].relation_created, self.reconcile)
         self.framework.observe(self.on["istio-pilot"].relation_changed, self.reconcile)
-        self.framework.observe(self.on["istio-pilot"].relation_broken, self.reconcile)
 
         # Watch:
         # * relation_changed: because if the remote data updates, we need to update our resources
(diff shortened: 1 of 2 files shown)


charms/istio-pilot/src/charm.py
```diff
@@ -450,6 +469,15 @@ def _get_gateway_service(self):
         )
         return svc
 
+    def _handle_istio_pilot_relation(self):
+        """Handles the istio-pilot relation, sending information about the Istio daemon."""
+        istio_pilot_interface = self._get_istio_pilot_interface()
+        if istio_pilot_interface:
+            # If something is related, send the data.  Otherwise skip
+            istio_pilot_interface.send_data(
+                {"service-name": f"istiod.{self.model.name}.svc", "service-port": "15012"}
```
beliaev-maksim, line 478: always the same port?

charms/istio-pilot/src/charm.py
```diff
@@ -115,10 +116,8 @@ def __init__(self, *args):
         # * relation_joined: because we send data to the other side whenever we see a related app
         # * relation_changed: because of SDI's data versioning model, which first agrees on the
         #                     schema version and then sends the rest of the data
-        # TODO: * relation_broken: is this needed?
         self.framework.observe(self.on["istio-pilot"].relation_created, self.reconcile)
         self.framework.observe(self.on["istio-pilot"].relation_changed, self.reconcile)
-        self.framework.observe(self.on["istio-pilot"].relation_broken, self.reconcile)
```
beliaev-maksim, line 121: rel_changed is not called on broken. Are we OK with it?

charms/istio-pilot/src/charm.py
ca-scribner: Yeah.  Currently we don't configure it in the istio install, so it is the default

charms/istio-pilot/src/charm.py
ca-scribner: Yeah, this relation is entirely an outward transmission of data.  We send out info about the istio daemon to whoever listens, and don't care if they stop listening.

beliaev-maksim (approved):

charms/istio-pilot/tests/unit/test_charm.py
```diff
@@ -336,6 +335,57 @@ def test_get_ingress_data_waiting_on_version(self, harness):
...
+            "service-port": "15012",
+        }
+
+        harness.set_leader(True)
+        harness.set_model_name(model_name)
+
+        relation_info = [
+            add_istio_pilot_to_harness(harness, other_app=name) for name in related_applications
+        ]
+        harness.begin()
```
beliaev-maksim, line 362: quick check, should not we harness begin before adding units?

I am 90% sure it should be before

beliaev-maksim (changes requested):
please check this internal resource: https://chat.canonical.com/canonical/pl/un6c4r3tp7f6bgtakuxm6nok4e

let me know what do you think

charms/istio-pilot/tests/unit/test_charm.py
ca-scribner: the correct way depends on what we want to test. the data is stored into the relation correctly in either order, but the charm reacts differently depending on the order:
1) begin, add relations, add data: when you add relations, relation created and joined events will fire. then same when you add data (relation changed event). 
2) add relations, add data, begin: no relation events will fire, and your charm will begin with data already available

These tests go with (2) because under test is just the helper, not the whole charm logic. If we do (1), we are testing that:
- we observed the relation event
- everything in the charm works successfully up until we get to what we are testing
- the method under test works however the charm wants to use it
- the rest of the charm also works

(1) is more an integration test, whereas (2) is more a unit test. If we try to use (1) here, we will have to mock a ton of stuff out to get the same result

beliaev-maksim (approved):

PR #253, opened 2023-04-21
feat: finish the istio-pilot main reconcile handler (for event handling rewrite)

ca-scribner:
This adds:
* handling for multiple non-fatal errors
* e2e-style unit tests using harness
* bugfixes for things found during unit testing

### DIFF
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -138,6 +138,13 @@ class Operator(CharmBase):
         # * relation_changed: because if the remote data updates, we need to update our resources
         # * relation_broken: because this is an application-level data exchange, so if the related
         #   application goes away we need to remove their resources
+        # This charm acts on relation_created/relation_joined to prevent accidentally leaving the
+        # ingress unsecured.  If anything is related to ingress-auth, that indicates the user wants
+        # block unauthenticated traffic through the Ingress, even if the related application has
+        # not yet sent us the authentication details.  By monitoring
+        # relation_created/relation_joined, we can block traffic through the Gateway until the auth
+        # is set up. This prevents a broken application on the ingress-auth relation from resulting
+        # in an unsecured ingress.
         self.framework.observe(self.on["ingress-auth"].relation_created, self.reconcile)
         self.framework.observe(self.on["ingress-auth"].relation_joined, self.reconcile)
         self.framework.observe(self.on["ingress-auth"].relation_changed, self.reconcile)
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -131,6 +131,15 @@ class Operator(CharmBase):
         #   application goes away we need to remove their resources
         self.framework.observe(self.on["ingress"].relation_changed, self.reconcile)
         self.framework.observe(self.on["ingress"].relation_broken, self.reconcile)
+
+        # Watch:
+        # * relation_created: because incomplete relation data should cause removal of the gateway
+        # * relation_joined: because incomplete relation data should cause removal of the gateway
+        # * relation_changed: because if the remote data updates, we need to update our resources
+        # * relation_broken: because this is an application-level data exchange, so if the related
+        #   application goes away we need to remove their resources
+        self.framework.observe(self.on["ingress-auth"].relation_created, self.reconcile)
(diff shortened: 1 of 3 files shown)


beliaev-maksim (approved):

charms/istio-pilot/src/charm.py
```diff
@@ -130,6 +131,15 @@ def __init__(self, *args):
         #   application goes away we need to remove their resources
         self.framework.observe(self.on["ingress"].relation_changed, self.reconcile)
         self.framework.observe(self.on["ingress"].relation_broken, self.reconcile)
+
+        # Watch:
+        # * relation_created: because incomplete relation data should cause removal of the gateway
+        # * relation_joined: because incomplete relation data should cause removal of the gateway
```
DnPlas, line 137: Is this right?

charms/istio-pilot/src/charm.py
```diff
@@ -380,12 +386,18 @@ def _get_interfaces(self):
...
-    def _get_ingress_auth_data(self) -> dict:
+    def _get_ingress_auth_data(self, event) -> dict:
         """Retrieve the ingress-auth relation data without touching other interface data.
 
         This is a workaround to ensure that errors in other relation data, such as an incomplete
         ingress relation, do not block us from retrieving the ingress-auth data.
         """
+        # Do not process data if this is a relation-broken event for this relation
+        if (
+            isinstance(event, RelationBrokenEvent)
```
DnPlas, line 397: Why not use observe and react to the relation broken event?

beliaev-maksim: I think I would agree with @DnPlas 

we can create another observer and put also this logic from `_reconcile_ingress_auth` there:
```python
        if len(ingress_auth_data) == 0:
            self.log.info("No ingress-auth data found - deleting any existing EnvoyFilter")
            _remove_envoyfilter(envoyfilter_name)
            _remove_envoyfilter(name=envoyfilter_name, namespace=self.model.name)
            return
```

i-chvets:
Doc strings in tests. Otherwise LGTM.

charms/istio-pilot/src/charm.py
beliaev-maksim: or no, since independent if we have ingress data or not we proceed to 
call  `_reconcile_gateway`

charms/istio-pilot/src/charm.py
ca-scribner: I think so - what part are you thinking seems off?  

The worry here is that if someone relates to ingress-auth but the application they relate with doesn't send the required data, the user would think have enabled authentication on the ingress when they haven't.  To avoid this, we react on any relation event and either:
* if the relation data is correct: create the envoyfilter and enforce the authentication
* if the relation data is incorrect or incomplete (maybe before the related app sends us data, or if the related app messes up and sends nothing), remove the gateway so there is no traffic through the ingress and report the error

This way at worst a misconfiguration blocks traffic rather than leaving something that should be secure insecure.  

What do you think?

charms/istio-pilot/src/charm.py
ca-scribner: We do observe the [`ingress-auth` relation_broken event](https://github.com/canonical/istio-operators/blob/4994fbf29bff048195e48cc55ee64995eebcbf16/charms/istio-pilot/src/charm.py#L144).  

When processing any relation_broken event, the relation data bag for that departing application *still exists and contains data*, so we cannot handle the event the same as other ingress-auth events.  If we don't treat this case special, it will still create the EnvoyFilter from the ingress-auth data even though the user is asking to turn that off.  

This is similar to [what we do for the `ingress` relation](https://github.com/canonical/istio-operators/blob/4994fbf29bff048195e48cc55ee64995eebcbf16/charms/istio-pilot/src/charm.py#L451), where we pop the breaking relation's data out of the data bag before processing.  But in the case of ingress-auth since it only supports being related to a single app, we can just return an empty dict instead.

charms/istio-pilot/src/charm.py
beliaev-maksim: @ca-scribner can you please add this explanation in the doc string? 
I believe that is very valuable for future us

ca-scribner: Added a comment to clarify this in the code.  This detail will also be in the design spec

beliaev-maksim (approved):

PR #256, opened 2023-04-23
fix: publishing CI in istio-pilot event handler feature branch

ca-scribner:
Cherry picks a bump in the charming-actions version from main.  This is to fix the publishing CI in the feature branch


i-chvets (approved):
Approved.

PR #263, opened 2023-04-26
fix: istio-pilot's handling of incomplete Gateway Services

ca-scribner:
This improves the handling of a gateway LoadBalancer service that does not have any ingress IPs.  Previously, this could fail if the ingress attribute was None, such as when there is no LoadBalancer provider present in the cluster.

This also adds a check during reconcile which will raise an ErrorWithStatus when the Gateway Service is not ready.

### DIFF
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -664,7 +664,7 @@ class Operator(CharmBase):
             )
             self._log_and_set_status(status_to_publish)
             for i, error in enumerate(errors):
-                self.log.info(f"Handled error {i}/{len(errors)}: {error.status}")
+                self.log.info(f"Handled error {i+1}/{len(errors)}: {error.status}")
         else:
             self.unit.status = ActiveStatus()
 
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -236,6 +236,12 @@ class Operator(CharmBase):
             handled_errors.append(err)
 
         try:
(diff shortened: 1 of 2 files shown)


charms/istio-pilot/src/charm.py
```diff
@@ -822,11 +839,13 @@ def _get_address_from_loadbalancer(svc):
           (str): The hostname or IP address of the LoadBalancer service
     """
     ingresses = svc.status.loadBalancer.ingress
+
+    # May happen due to the Kubernetes cluster not having a LoadBalancer provider
+    if ingresses is None or len(ingresses) == 0:
```
beliaev-maksim, line 844: ?
```suggestion
    if not ingresses:
```

charms/istio-pilot/src/charm.py
```diff
@@ -647,7 +664,7 @@ def _report_handled_errors(self, errors):
             )
             self._log_and_set_status(status_to_publish)
             for i, error in enumerate(errors):
-                self.log.info(f"Handled error {i}/{len(errors)}: {error.status}")
+                self.log.info(f"Handled error {i+1}/{len(errors)}: {error.status}")
```
beliaev-maksim, line 667: you can use `enumerate(errors, start=1)`
```suggestion
                self.log.info(f"Handled error {i}/{len(errors)}: {error.status}")
```

beliaev-maksim (changes requested):
also, I just think about `_report_handled_errors`. Do we really handle them ?

PR #272, opened 2023-05-23
Downgrade juju

beliaev-maksim:
we are on 2.9, thus, pylibjuju should be 2.9


ca-scribner (changes requested):
I don't believe this PR is correct.  pylibjuju 3.0.x is supposed to be compatible with 2.9 and 3.x I believe, but also gets better patch updates than 2.9.x.  So I think it is the version we want in this case.  

I might be wrong about this though, I've been confused on which we need.  Was there something prompting this?

beliaev-maksim:
@ca-scribner I talked with Juan

3.0 is not maintained. All bug fixes are landed in 2.9, 3.1 and 3.2

ca-scribner:
ok sounds good to me then.  thanks!

ca-scribner (approved):

PR #275, opened 2023-05-25
KF-3009 update k8s api version

beliaev-maksim:
change API version.
safe change since this is supported in K8s since 1.21

https://kubernetes.io/docs/reference/using-api/deprecation-guide/#poddisruptionbudget-v125

once updated on local charm within a bundle the resource can be successfully applied

image


i-chvets (approved):
Approved.

PR #277, opened 2023-05-29
Fix handling of SSL

ca-scribner:
This PR updates the EnvoyFilter definition to always apply auth to both http and https traffic, regardless of which is currently enabled in the Gateway.  This is to prevent the possibility of these objects going out of sync (eg: Gateway allowing traffic on one port, and EnvoyFilter applying Auth on traffic through the other port), which could have occurred if the charm succeeded in updating one of those objects but failed to update the other (due to an error or incompatible settings).  By always applying the EnvoyFilter to all traffic, this possibility is prevented.

This PR came from an attempt to address https://github.com/canonical/bundle-kubeflow/issues/570, which showed our previously was misconfiguring the `EnvoyFilter` when using SSL.  It was reported that we incorrectly applied the `EnvoyFilter` only to http traffic and not https traffic.  Reproducing https://github.com/canonical/bundle-kubeflow/issues/570 was a challenge, partly because the charm has gone through recent rewrites that might have fixed the issue.  The specific cause of https://github.com/canonical/bundle-kubeflow/issues/570 could not be confirmed.  It may be fixed by this PR, but that is unclear.  More discussion on this is in [this thread](https://github.com/canonical/istio-operators/pull/277#discussion_r1214646654).

Closes https://github.com/canonical/bundle-kubeflow/issues/570

### Testing instructions

```bash
# pack the modified charm (from this PR)
cd charms/istio-pilot
charmcraft pack

# deploy istio, dex/oidc, and the dashboard to have something to browse to.  
# For the initial deployment, do this without certs/using http
juju deploy ./istio-pilot_ubuntu-20.04-amd64.charm --trust
juju deploy istio-gateway --channel=1.15/stable --trust --config kind=ingress istio-ingressgateway

juju deploy dex-auth --channel=2.31/stable --trust --config static-username=user --config static-password=user --config public-url=http://10.64.140.43.nip.io
juju deploy oidc-gatekeeper --channel ckf-1.7/stable --config public-url=http://10.64.140.43.nip.io
juju relate istio-pilot:ingress dex-auth:ingress
juju relate dex-auth:oidc-client oidc-gatekeeper:oidc-client
juju relate istio-pilot:ingress oidc-gatekeeper:ingress
juju relate istio-pilot:ingress-auth oidc-gatekeeper:ingress-auth

juju deploy kubeflow-profiles --channel=1.7/stable --trust
juju deploy kubeflow-dashboard --channel=1.7/stable --trust
juju relate kubeflow-profiles kubeflow-dashboard
juju relate kubeflow-dashboard:ingress istio-pilot:ingress

# Check whether everything works as expected
# try to connect to 
# * http://10.64.140.43.nip.io/: redirected to dex login 
# * http://10.64.140.43.nip.io/dex/.well-known/openid-configuration: works
# * https://10.64.140.43.nip.io/: cannot be reached
# * https://10.64.140.43.nip.io/dex/.well-known/openid-configuration: cannot be reached
# http all works, https all does not, as expected

# Set up SSL by creating a self-signed cert.  Note:
# * if you used a real cert, you wouldn't have to update the oidc-gatekeeper ca-bundle. 
# * you must have a common name and SAN in the self-signed cert

openssl req -nodes -new -x509 \
 -subj "/CN=10.64.140.43.nip.io" \
 -addext "subjectAltName = DNS:10.64.140.43.nip.io" \
 -keyout server-san.key -out server-san.cert 

juju config istio-pilot ssl-crt="$(cat server-san.cert | base64 -w0)"
juju config istio-pilot ssl-key="$(cat server-san.key | base64 -w0)" 

juju config dex-auth public-url=https://10.64.140.43.nip.io

# Because oidc for 1.7 does not restart its workload if you change the ca-bundle, we remove and redeploy it
juju remove-application oidc-gatekeeper
juju deploy oidc-gatekeeper --channel ckf-1.7/stable --config public-url=https://10.64.140.43.nip.io --config ca-bundle="$(cat server-san.cert)"

# Re-relate oidc
juju relate dex-auth:oidc-client oidc-gatekeeper:oidc-client
juju relate istio-pilot:ingress oidc-gatekeeper:ingress
juju relate istio-pilot:ingress-auth oidc-gatekeeper:ingress-auth

# wait for oidc/dex-auth to do their thing...

# Check whether everything works.  Now, we should only be able to reach http urls, not https
# * http://10.64.140.43.nip.io/: cannot be reached
# * http://10.64.140.43.nip.io/dex/.well-known/openid-configuration: cannot be reached
# * https://10.64.140.43.nip.io: warning about insecure cert, then redirected to dex login
# * https://10.64.140.43.nip.io/dex/.well-known/openid-configuration: warning about insecure cert, then it works
```

### DIFF
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -613,7 +613,7 @@ class Operator(CharmBase):
             "app_name": self.app.name,
             "envoyfilter_name": envoyfilter_name,
             "envoyfilter_namespace": self.model.name,
-            "gateway_port": self._gateway_port,
+            "gateway_ports": list(GATEWAY_PORTS.values()),
             "port": ingress_auth_data["port"],
             "request_headers": ingress_auth_data["allowed-request-headers"],
             "response_headers": ingress_auth_data["allowed-response-headers"],
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -48,8 +48,10 @@ VIRTUAL_SERVICE_LIGHTKUBE_RESOURCE = create_namespaced_resource(
     plural="virtualservices",
 )
 
-GATEWAY_HTTP_PORT = 8080
-GATEWAY_HTTPS_PORT = 8443
+GATEWAY_PORTS = {
+    "http": 8080,
+    "https": 8443,
+}
 GATEWAY_TEMPLATE_FILES = ["src/manifests/gateway.yaml.j2"]
 KRH_GATEWAY_SCOPE = "gateway"
 METRICS_PORT = 15014
@@ -369,9 +371,9 @@ class Operator(CharmBase):
     @property
     def _gateway_port(self):
         if self._use_https():
-            return GATEWAY_HTTPS_PORT
+            return GATEWAY_PORTS["https"]
         else:
-            return GATEWAY_HTTP_PORT
+            return GATEWAY_PORTS["http"]
 
     def _get_interfaces(self):
         """Retrieve interface object."""
--- a/charms/istio-pilot/src/manifests/auth_filter.yaml.j2
+++ b/charms/istio-pilot/src/manifests/auth_filter.yaml.j2
@@ -6,6 +6,7 @@ metadata:
   namespace: {{ envoyfilter_namespace }}
 spec:
   configPatches:
+  {% for gateway_port in gateway_ports %}
     - applyTo: HTTP_FILTER
       match:
         context: GATEWAY
@@ -39,6 +40,7 @@ spec:
                       {% for response_header in response_headers %}
                       - exact: {{ response_header }}
                       {% endfor %}
+  {% endfor %}
 
   workloadSelector:
     labels:
--- a/charms/istio-pilot/tests/unit/test_charm.py
+++ b/charms/istio-pilot/tests/unit/test_charm.py
@@ -30,8 +30,7 @@ from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingSta
 from ops.testing import Harness
 
 from charm import (
-    GATEWAY_HTTP_PORT,
-    GATEWAY_HTTPS_PORT,
+    GATEWAY_PORTS,
     Operator,
     _get_gateway_address_from_svc,
     _remove_envoyfilter,
@@ -435,8 +434,8 @@ class TestCharmHelpers:
     @pytest.mark.parametrize(
         "ssl_crt, ssl_key, expected_port, expected_context",
         [
-            ("", "", GATEWAY_HTTP_PORT, does_not_raise()),
-            ("x", "x", GATEWAY_HTTPS_PORT, does_not_raise()),
+            ("", "", GATEWAY_PORTS["http"], does_not_raise()),
+            ("x", "x", GATEWAY_PORTS["https"], does_not_raise()),
             ("x", "", None, pytest.raises(ErrorWithStatus)),
             ("", "x", None, pytest.raises(ErrorWithStatus)),
         ],
--- a/charms/istio-pilot/tests/unit/test_charm.py
+++ b/charms/istio-pilot/tests/unit/test_charm.py
@@ -1480,13 +1480,33 @@ def is_lightkube_resource_in_call_args_list(call_args_list, name, namespace=None
         namespace (str): The namespace of the lightkube resource to search for (will check
                          metadata.namespace)
     """
+    try:
+        get_lightkube_resource_in_call_args_list(call_args_list=call_args_list, name=name, namespace=namespace)
+        return True
+    except KeyError:
+        return False
+
+
+def get_lightkube_resource_in_call_args_list(call_args_list, name, namespace=None):
+    """Returns the first call_args from a call_args_list that was for a given lightkube resource.
+
+    Raises a KeyError if the object does not exist.
+
+    Args:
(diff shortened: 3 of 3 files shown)


charms/istio-pilot/src/charm.py
```diff
@@ -48,8 +48,10 @@
     plural="virtualservices",
 )
 
-GATEWAY_HTTP_PORT = 8080
-GATEWAY_HTTPS_PORT = 8443
+GATEWAY_PORTS = {
+    "http": 8080,
+    "https": 8443,
+}
```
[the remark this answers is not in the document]
ca-scribner: Yeah during the refactor it felt a bit more natural this way, but either works.

charms/istio-pilot/src/manifests/auth_filter.yaml.j2
```diff
@@ -39,6 +40,7 @@ spec:
                       {% for response_header in response_headers %}
```
DnPlas, line 40: You already have a `{% for response_header in response_headers %}` loop in L34, why don't you re-use it?

charms/istio-pilot/src/manifests/auth_filter.yaml.j2
```diff
@@ -6,6 +6,7 @@ metadata:
   namespace: {{ envoyfilter_namespace }}
 spec:
   configPatches:
+  {% for gateway_port in gateway_ports %}
```
DnPlas, line 9: Is it possible for `portNumber` to have more than one value? How would this look like when rendered?

charms/istio-pilot/src/charm.py
```diff
@@ -611,7 +613,7 @@ def _reconcile_ingress_auth(self, ingress_auth_data: dict):
             "app_name": self.app.name,
             "envoyfilter_name": envoyfilter_name,
             "envoyfilter_namespace": self.model.name,
-            "gateway_port": self._gateway_port,
+            "gateway_ports": list(GATEWAY_PORTS.values()),
```
DnPlas, line 616: What I understand from your PR is that you are passing a list of gateway ports as context to render auth filter manifest file so the things inside the for loop are rendered twice, that way you are applying the EnvoyFilter to both http and https traffic no matter the charm configuration. That being said and just for my own understanding, why do we want to apply the EnvoyFilter to both http and https traffic?

DnPlas (commented):
Thanks @ca-scribner ! Some comments.

i-chvets:
My only comment would be to move testing insructions into related issue instead of PR.

ca-scribner:
Good call @i-chvets - I'll make an issue, add the steps, and close the issue.

charms/istio-pilot/src/manifests/auth_filter.yaml.j2
ca-scribner: ohh this is an easy one to miss.  line 34 loops on `request_headers`, whereas this is `response_headers`

charms/istio-pilot/src/manifests/auth_filter.yaml.j2
ca-scribner: I think [this conversation](https://github.com/canonical/istio-operators/pull/277#discussion_r1212891463) will apply here too

charms/istio-pilot/src/charm.py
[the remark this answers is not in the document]
beliaev-maksim: do not know, I am -0 on the change :D

up to you :)

charms/istio-pilot/src/manifests/auth_filter.yaml.j2
DnPlas: Oh, even in the comment I had a hard time. Okay!

charms/istio-pilot/src/charm.py
DnPlas: > Am I understanding that correctly?

Yes

> Purely out of fear.

If this fix is out of pure fear, then [this](https://discourse.charmhub.io/t/kubeflow-1-7-ssl-and-dex-auth/9574) issue is already fixed in the existing code (before this PR), isn't?

What I am trying to understand is why we need this. In the charm (before this PR), we are supposedly covering both cases, not at the same time, but we know both cases should be covered.

ca-scribner: After some offline discussions with @DnPlas and @kimwnasptd, we will merge this as is because it is at worst helpful even if it didn't solve the bug, but then I'll dig further to understand exactly what happened separately.

ca-scribner: After some additional testing, I can confirm

* istio-pilot shipped with Kubeflow 1.6: works as expected
* istio-pilot shipped with Kubeflow 1.7: does not work
* istio-pilot refactored [here](https://github.com/canonical/istio-operators/pull/261) after 1.7: works as expected
* this PR: works as expected

So https://github.com/canonical/bundle-kubeflow/issues/570 was really closed by https://github.com/canonical/istio-operators/pull/261, and this PR adds a bit of safety that prevents a possible problem in future (but one that we've never hit in practice that we know of).

DnPlas: Could you please add your findings in the description of this issue (or at least a link to your previous comments) so everyone can see them once this PR is merged?

ca-scribner: yeah good call.  I'll add a link to this so the whole conversation is easy to spot

DnPlas (approved):

DnPlas:
Thanks @ca-scribner, LGTM

PR #286, opened 2023-06-06
fix: fix flaky istio-pilot unit tests

ca-scribner:
Updates the istio-pilot Gateway unit tests to handle lightkube being called in a non-determinstic order.  Previously, we asserted that two calls happened in a particular order but this is not guaranteed.  Now the tests assert only that the two calls happen, as order does not matter.


misohu (approved):
If the tests pass LGTM

DnPlas (approved):

ca-scribner:
Looks like the test failure was a network flake during publishing.  Merging anyway

Issue #189, opened 2023-03-13
istio-pilot fails to process `ingress-relation-broken` event during charm remove


ca-scribner:
`juju remove-application istio-pilot` will fail with the following debug-log:

```
Traceback (most recent call last):
  File "./src/charm.py", line 430, in <module>
    main(Operator)
  File "/var/lib/juju/agents/unit-istio-pilot-0/charm/venv/ops/main.py", line 438, in main
    _emit_charm_event(charm, dispatcher.event_name)
  File "/var/lib/juju/agents/unit-istio-pilot-0/charm/venv/ops/main.py", line 150, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-istio-pilot-0/charm/venv/ops/framework.py", line 355, in emit
    framework._emit(event)  # noqa
  File "/var/lib/juju/agents/unit-istio-pilot-0/charm/venv/ops/framework.py", line 856, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-istio-pilot-0/charm/venv/ops/framework.py", line 931, in _reemit
    custom_handler(event)
  File "./src/charm.py", line 257, in handle_ingress
    del routes[(event.relation, event.app)]
KeyError: (<ops.model.Relation ingress:3>, None)
2023-03-13 18:56:53 ERROR juju.worker.uniter.operation runhook.go:140 hook "ingress-relation-broken" (via hook dispatching script: dispatch) failed: exit status 1
2023-03-13 18:56:53 INFO juju.worker.uniter resolver.go:145 awaiting error resolution for "relation-broken" hook
2023-03-13 18:58:39 INFO juju.worker.uniter resolver.go:145 awaiting error resolution for "relation-broken" hook
```

It appears that [this line](https://github.com/canonical/istio-operators/blob/main/charms/istio-pilot/src/charm.py#L257) does not work correctly.  

I have not tested it, but this probably affects the ingress-relation-broken event independent of it being called during a remove-application

ca-scribner:
Closed by #261

ca-scribner:
I didn't hit this when working on #277, but @NohaIhab later hit it when following those steps for something unrelated.  I wonder if there's a race condition at play here?  Maybe if the other app is already gone, `event.app` will show `None` instead of the app name? 

We could try to fix this by skipping the `routes.pop` if `event.app is None`, but it isn't clear whether the data in the relation is actually gone at this point or not.

[...]

PR #291, opened 2023-06-14
fix: handle ingress `relation-broken` when `event.app==None`

ca-scribner:
While handling relation-broken events for ingress, we see either:
* event.app=the_departing_app
* event.app=None

The previous charm logic did not cover the second case, leading to an uncaught KeyError.

We aren't sure what controls which of the situations we encounter, but we add error handling here to at least log the event and avoid uncaught exceptions.

Fixes https://github.com/canonical/istio-operators/issues/189
Part of #292

### DIFF
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -447,10 +447,18 @@ class Operator(CharmBase):
         # Get all route data from the ingress interface
         routes = get_routes_from_ingress_interface(ingress_interface, self.app)
 
-        # The app-level data is still visible on a broken relation, but we
+        # The app-level data may still be visible on a broken relation, but we
         # shouldn't be keeping the VirtualService for that related app.
         if isinstance(event, RelationBrokenEvent) and event.relation.name == INGRESS_RELATION_NAME:
-            routes.pop((event.relation, event.app))
+            if event.app is None:
+                self.log.info(
+                    f"When handling event '{event}', event.app found to be None.  We cannot pop"
+                    f" the departing application's data from 'routes' because we do not know the"
+                    f" departing application's name.  We assume that the departing application's"
+                    f" is not in routes.keys='{list(routes.keys())}'."
+                )
+            else:
+                routes.pop((event.relation, event.app))
 
         return routes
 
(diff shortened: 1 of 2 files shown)


charms/istio-pilot/src/charm.py
```diff
@@ -447,10 +447,18 @@ def _get_ingress_data(self, event) -> dict:
         # Get all route data from the ingress interface
         routes = get_routes_from_ingress_interface(ingress_interface, self.app)
 
-        # The app-level data is still visible on a broken relation, but we
+        # The app-level data may still be visible on a broken relation, but we
         # shouldn't be keeping the VirtualService for that related app.
```
DnPlas, line 451: Do you mind adding a comment here that says that the VS for the removed remote application could still be present in the cluster and what could happen with it?

DnPlas (commented):
Thanks @ca-scribner, just a tiny comment, other than that LGTM.

i-chvets:
LGTM. Ci needs to pass.

charms/istio-pilot/src/charm.py
```diff
@@ -431,10 +431,18 @@ def _get_ingress_data(self, event) -> dict:
...
         if isinstance(event, RelationBrokenEvent) and event.relation.name == INGRESS_RELATION_NAME:
-            routes.pop((event.relation, event.app))
+            if event.app is None:
+                self.log.info(
```
i-chvets, line 438: Should this be an ERROR log level and not INFO?

charms/istio-pilot/tests/unit/test_charm.py
```diff
@@ -591,6 +591,21 @@ def test_get_ingress_data_for_broken_event(self, harness):
...
+        # above relations.
+        mock_relation_broken_event = MagicMock(spec=RelationBrokenEvent)
+        mock_relation_broken_event.relation = harness.model.get_relation("ingress", 1)
+        mock_relation_broken_event.app = None
+
+        # Mock the logger
+        harness.charm.log = MagicMock()
+
+        harness.charm._get_ingress_data(mock_relation_broken_event)
+        assert harness.charm.log.info.call_count == 1
```
i-chvets, line 607: If INFO in the code is changed to ERROR, this will need to change as well.

charms/istio-pilot/src/charm.py
DnPlas: We want just to log the incident and move on.

i-chvets (approved):
Approved.

PR #310, opened 2023-08-16
ci: use matrix strategy to test on different microk8s versions

DnPlas:
Using a matrix to test the istio-operators in the two supported versions of microk8s (1.25 and 1.26). Also installing the strict version of microk8s in preparation for juju 3.1 usage.


i-chvets (approved):
Approved.

PR #312, opened 2023-08-16
chore: remove patches for istio < 1.16

DnPlas:
Istio 1.12-1.14 had an issue with the ValidatingWebhookConfiguration and the namespace, which we patched in the charm code to prevent errros. Since this issue is not present in recent versions of the charm, we do not need the workaround anymore. This code is still valid in previous versions and can be left in those branches.


i-chvets (approved):
Approved.

Issue #320, opened 2023-08-29
Istio-pilot busybox image is hardcoded


i-chvets:
# Description
Currently, compiled manifests in `istioctl` do not allow modificatrion of `grpc-simple.initContainers.image`. It is hardcoded to `busybox:xx` and there is no parameter exposed to be altered via `--set` argument.

Prelimnary work is done in https://github.com/canonical/istio-operators/tree/KF-3996-control-plane-configurable-images and was aiming to install rendered manifests with `busybox` replaced with configuration value. However, this requires enabling [storage](https://juju.is/docs/sdk/storage) for the charm. 

Alternative soltion: Ignore hardcoded `busybox` image and use `--set` to modify required paramters. Test out this in arigapped enviroment to verify if `busybox` is indeed required. This solution does not requires storage, because `istioctl` binary will iuse pre-compiled manifests whose parameters will be adjusted via `--set`.

i-chvets:
There is no feasible way to set `busybox` image when deploying `istio-pilot`
The following PR #321 will only enable configuratiobn of `istio-pilot` and `istio-proxy` images.

PR #321, opened 2023-08-29
Kf 3996 control plane configurable images with set

i-chvets:
More details in: https://github.com/canonical/istio-operators/issues/320

Summary of changes:
- Modified code to use --set to alter precompile istioctl manifests to allow changes according to charm configuration.

NOTE: This is en experimental PR for now.


charms/istio-pilot/config.yaml
```diff
@@ -3,6 +3,18 @@ options:
...
     default: istio-gateway
     description: Name to use as a default gateway
+  image-configuration:
+    default: |
+      pilot-image: 'pilot'  # values.pilot.image
+      global-tag: '1.17.3'  # values.global.tag
+      global-hub: 'docker.io/istio' # values.global.hub
+      global-proxy-image: 'proxyv2' # values.global.proxy.image
+      global-proxy-init-image: 'proxyv2' # values.global.proxy_init.image
+      grpc-bootstrap-init: 'busybox:1.28'
```
orfeas-k, line 13: Where do those values come from? I can see the referenced doc but it is for another version and has different values at some places. This brings us to the next question as to how could one update those values in the case of bumping istio version for a new CKF release? Maybe, we will need to update more than just `global-tag`
Maybe we can proceed as it is now, but we 'll definitely need to tackle this soon.

orfeas-k (approved):

PR #339, opened 2023-10-02
docs: add TLS certificates integration instructions

DnPlas:
Adds instructions for enabling TLS ingress gateway on the istio control plane using a TLS certificates provider.

Merge after #338

### DIFF
--- a/charms/istio-pilot/README.md
+++ b/charms/istio-pilot/README.md
@@ -11,6 +11,21 @@ juju deploy istio-pilot --trust
 juju deploy istio-gateway --trust --config kind=ingress istio-ingressgateway
 juju relate istio-pilot istio-ingressgateway
 ```
+## Enable TLS ingress gateway for a single host
+
+This charm provides means to integrate with TLS certificates providers that help with this configuration. The following can be used as TLS certificates providers depending on the use case or security requirements of an organisation:
+
+* For development or non-production environments, the istio-pilot charm can be related to the [self-signed-certificates-operator](https://github.com/canonical/self-signed-certificates-operator/tree/main).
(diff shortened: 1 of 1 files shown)


charms/istio-pilot/README.md
```diff
@@ -11,6 +11,21 @@ juju deploy istio-pilot --trust
 juju deploy istio-gateway --trust --config kind=ingress istio-ingressgateway
 juju relate istio-pilot istio-ingressgateway
 ```
+## Enable TLS ingress gateway for a single host
+
+This charm provides means to integrate with TLS certificates providers that help with this configuration. The following can be used as TLS certificates providers depending on the use case or security requirements of an organisation:
+
+* For development or non-production environments, the istio-pilot charm can be related to the [self-signed-certificates-operator](https://github.com/canonical/self-signed-certificates-operator/tree/main).
+* For environments where certificates are managed by the organisation, the [tls-certificates-operator](https://github.com/canonical/tls-certificates-operator) can be used.
+* For deployments where it is desired to use ACME servers for validation (like Let's encrypt), you can use any of `httpreq-acme-operator`, `namecheap-acme-operator`, or [route53-acme-operator](https://github.com/canonical/route53-acme-operator).
```
orfeas-k, line 20: I think we should have links for all acme operators, unless there's a reason not to.

orfeas-k (commented):
Left a minor comment, other than that looks good. Thank you @DnPlas

charms/istio-pilot/README.md
DnPlas: Updated!

orfeas-k (approved):

Issue #340, opened 2023-10-02
istio-pilot won't build on CI


DnPlas:
### Bug Description

The changes introduced by #338 are preventing the charm from building. This PR is introducing new build packages that require running `apt update`, but the process fails with `Permission denied` messages. 

### To Reproduce

1. Run the CI in #338 

### Environment

Ubuntu 20.04
juju 3.1
microk8s 1.25-strict/stable

### Relevant log output

Logs [here](https://github.com/canonical/istio-operators/actions/runs/6378869306/job/17311184751#step:6:1)

### Additional context

~~I haven't been able to reproduce the issue locally, but the charm just doesn't seem to build at all, it gets stuck.~~

Update on reproducing locally: I don't see any error in a local testing environment, I was able to run every test case just fine.

PR #341, opened 2023-10-02
ci: run integration tests w/o destructive mode

DnPlas:
As canonical/charmcraft#1132 got fixed, destructive mode can now be removed from the tox command when running integration tests. The fix is in latest/edge, this commit also points to that version.
minor nit: fix the path of charmcraft logs for collecting them.

Fixes: #340


i-chvets (approved):
Approved.

PR #342, opened 2023-10-06
fix: set charm to waitingStatus if missing ingress auth data

DnPlas:
Put istio-pilot into waitingStatus until the ingress-auth data provider changes the data in the relation data bag.

Fixes #336 
Fixes #346

PR #348, opened 2023-11-21
fix(cherry-pick): set charm to waitingStatus if missing ingress auth data (#342)

orfeas-k:
* fix: set charm to waitingStatus if missing ingress auth data
* Add unit tests


DnPlas (approved):
Thanks @orfeas-k LGTM

PR #349, opened 2023-11-21
Use fastapi automation for jira sync

kimwnasptd:
change the automation to instead use https://github.com/canonical/gh-jira-sync-bot for creating issues, instead of the previous workflow


DnPlas (approved):
Thanks @kimwnasptd LGTM

Issue #287, opened 2023-06-07
istio-pilot deployment fails when there is no LoadBalancer in the cluster


orfeas-k:
### Context
Trying to run [this test](https://github.com/canonical/kubeflow-tensorboards-operator/blob/main/tests/integration/test_charms.py) for [kubeflow-tensorboards-operator](https://github.com/canonical/kubeflow-tensorboards-operator) (which is a simple "deploy and relate the minimum required charms" test), we noticed that istio-pilot charm would fail in different parts of its execution. The error always looked like the one below, but the hook failing could also be `install`, `istio-pilot-relation-changed`. 
```
...
  File "./src/charm.py", line 827, in _get_address_from_loadbalancer
    if len(ingresses) != 1:
TypeError: object of type 'NoneType' has no len()
unit-istio-pilot-0: 15:19:55 ERROR juju.worker.uniter.operation hook "config-changed" (via hook dispatching script: dispatch) failed: exit status 1
```

### Reproduce
 - [Install and setup Microk8s](https://juju.is/docs/olm/microk8s#heading--install-and-set-up-microk8s) 
```
sudo snap install microk8s --classic --channel=1.24/stable
sudo usermod -a -G microk8s $USER
sudo chown -f -R $USER ~/.kube
microk8s status --wait-ready
microk8s.enable hostpath-storage dns 
sudo snap install juju --classic
juju bootstrap microk8s micro
```
- `git clone git@github.com:canonical/kubeflow-tensorboards-operator.git` and  `cd kubeflow-tensorboards-operator` 
- Install [test requirements](https://github.com/canonical/kubeflow-tensorboards-operator/blob/main/requirements-integration.txt) `python3 -m pip install -r requirements-integration.txt`
- Run test with `tox -e integration` 

### Cause
It turns out the `Microk8s` cluster didn't have a LoadBalancer and Istio relies heavily on the underlying LB to configure the gateway (ingressgateway).  Otherwise, the gateway will be waiting for an external IP and `istio-pilot` will fail due to the [following lines](https://github.com/canonical/istio-operators/blob/38ce4cb413a5d7c9184a5e36a91cdfded5213048/charms/istio-pilot/src/charm.py#L826-L831) where it checks the `ingresses` object's length before verifying that it is not `None`. Probably though, this is a case that the `istio-pilot` charm should be able to handle. 

P.S. The error was gone as soon as I added a LB with `microk8s enable metallb:10.64.140.43-10.64.140.49`

[...]

PR #353, opened 2023-11-24
Handle None being returned for ingresses

motjuste:
`svc.status.loadBalancer.ingress` has been observed to return `None` in certain situations and leads to failure of the `ingress-relation-created` hook.  This change handles the `None` and treats the situation as if no ingresses are present.

### DIFF
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -827,11 +827,10 @@ def _get_address_from_loadbalancer(svc):
           (str): The hostname or IP address of the LoadBalancer service
     """
     ingresses = svc.status.loadBalancer.ingress
-    if len(ingresses) != 1:
-        if len(ingresses) == 0:
-            return None
-        else:
-            raise ValueError("Unknown situation - LoadBalancer service has more than one ingress")
+    if ingresses is None or len(ingresses) == 0:
+        return None
+    elif len(ingresses) != 1:
+        raise ValueError("Unknown situation - LoadBalancer service has more than one ingress")
 
     ingress = svc.status.loadBalancer.ingress[0]
     if getattr(ingress, "hostname", None) is not None:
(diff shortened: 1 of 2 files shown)


motjuste:
I am happy to write tests for this, but would need guidance to get started.

orfeas-k:
Hi @motjuste, does this PR try to resolve https://github.com/canonical/istio-operators/issues/287?

motjuste:
> Hi @motjuste, does this PR try to resolve #287?

Not intentionally, but yes, I also faced the same issue.

orfeas-k:
That's great. Thank you for taking the time to try and fix this. Could you add a unit test then that replicates the situtation as described in the issue https://github.com/canonical/istio-operators/issues/287#issuecomment-1581196275? This way, we 'll make sure that the fix you 've sent makes what we expect it to.

You could do that by adding another [fixture](https://github.com/canonical/istio-operators/blob/main/charms/istio-pilot/tests/unit/test_charm.py#L1426-L1436) for the specific case and then adding it as a parameter to the tests we already have ([first](https://github.com/canonical/istio-operators/blob/main/charms/istio-pilot/tests/unit/test_charm.py#L535-L546) and [second](https://github.com/canonical/istio-operators/blob/main/charms/istio-pilot/tests/unit/test_charm.py#L506-L517)).

motjuste:
Thanks for pointing me to the tests.  I'll update the PR with them.

motjuste:
Added a new mock fixture (sorry for the verbose name, suggestions welcome), and then I added it to the two parameterised tests.  

I can't seem to check the CI outputs.  I am new to GitHub Actions ... is there a way for me to get the test running in GitHub without having to set everything up on my laptop?

orfeas-k:
Hi @motjuste, we discussed this issue with our team internally and decided that we should start allowing PRs from external forks to run workflows (issue https://github.com/canonical/bundle-kubeflow/issues/764). We will look to implement this in the near future but at the moment, this is not possible (for security reasons as mentioned in the issue). 

For that reason, we will hold this PR until that issue is resolved, so we can properly test and merge it. Once again, thank you for taking the time to contribute to our project.

Issue #351, opened 2023-11-21
Install Istio control plane with CNI plugin using istio-pilot


DnPlas:
### What needs to get done

Based on the implementation details from the spec (https://github.com/canonical/istio-operators/issues/350), install the Istio control plane with the CNI plugin. This task requires changes in the `istio-pilot` charm code, potentially:

* Using our [install method](https://github.com/canonical/istio-operators/blob/main/charms/istio-pilot/src/charm.py#L190C1-L211C10), enable the plugin using setting a new `values.istio_cni.enabled` 
* Ensuring the istio-pilot charm can be upgraded from a non-cni installation

### Why it needs to get done

Some changes need to be done on the `istio-pilot` charm code to be able to install the Istio CNI plugin.

DnPlas:
Fixed by #365

PR #357, opened 2023-12-04
refactor: use the istioctl wrapper for installing the control plane

DnPlas:
This refactor uses the install method form the istioctl wrapper to install the control plane without calling subprocess directly in the charm code.

Part of #351


charms/istio-pilot/src/charm.py
```diff
@@ -66,6 +65,11 @@
 VIRTUAL_SERVICE_TEMPLATE_FILES = ["src/manifests/virtual_service.yaml.j2"]
 ISTIOCTL_PATH = "./istioctl"
 ISTIOCTL_DEPOYMENT_PROFILE = "minimal"
+INSTALL_FAILED_MSG = (
+    "Failed to install Istio Control Plane."
```
ca-scribner, line 69: Missing whitespace (and I didn't even put two spaces here because I know you prefer one.  Even though two is definitely better :) )
```suggestion
    "Failed to install Istio Control Plane. "
```

charms/istio-pilot/src/charm.py
```diff
@@ -187,29 +192,36 @@ def install(self, _):
...
+            ISTIOCTL_PATH,
+            self.model.name,
+            ISTIOCTL_DEPOYMENT_PROFILE,
+            istioctl_extra_flags=istioctl_extra_flags,
         )
 
+        try:
+            istioctl.install()
+        except IstioctlError as e:
+            self.log.error(INSTALL_FAILED_MSG.format(message="istioctl install failed"))
```
ca-scribner, line 220: Any reason why we don't use `message=str(e)` so we provide the error message that Istioctl creates?  I think it has a bit more information, and is the same as how other errors are passed to the user

charms/istio-pilot/src/istioctl.py
```diff
@@ -11,34 +12,50 @@ class IstioctlError(Exception):
 
 class Istioctl:
     def __init__(
-        self, istioctl_path: str, namespace: str = "istio-system", profile: str = "minimal"
+        self,
+        istioctl_path: str,
+        namespace: str = "istio-system",
+        profile: str = "minimal",
+        istioctl_extra_flags: Optional[List[str]] = [],
```
ca-scribner, line 19: Initialising this to `[]` has the unexpected side effect of sharing the default value between all instances of the class ([ref 1](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments), [ref 2](https://stackoverflow.com/questions/4841782/python-constructor-and-default-value)).  Can we change this to `= None` and then have an `if istioctl_extra_flags = None` in the init?
 an unexpected side effect because `[]` is mutable.  It basically means all instances of `Istioctl` will share

charms/istio-pilot/src/istioctl.py
```diff
@@ -11,34 +12,50 @@ class IstioctlError(Exception):
...
         """Wrapper for the istioctl binary.
 
         This class provides a python API for the istioctl binary, supporting install, upgrade,
         and other istioctl commands.
 
         Args:
+            istioctl_extra_flags (optional, list): A list containing extra flags to pass to istioctl
             istioctl_path (str): Path to the istioctl binary to be wrapped
             namespace (str): The namespace to install Istio into
             profile (str): The Istio profile to use for installation or upgrades
```
ca-scribner, line 30: Can we move this to the bottom of the list to keep it in the same order as the flags?

```suggestion
            istioctl_path (str): Path to the istioctl binary to be wrapped
            namespace (str): The namespace to install Istio into
            profile (str): The Istio profile to use for installation or upgrades
            istioctl_extra_flags (optional, list): A list containing extra flags to pass to istioctl
```

charms/istio-pilot/tests/unit/test_istioctl.py
```diff
@@ -64,11 +70,32 @@ def test_istioctl_install(mocked_check_call):
...
-        f"profile={PROFILE}",
-        "-s",
-        f"values.global.istioNamespace={NAMESPACE}",
     ]
+    expected_call_args.extend(EXPECTED_ISTIOCTL_FLAGS)
+
+    mocked_check_call.assert_called_once_with(expected_call_args)
+
+
+def test_istioctl_install_extra_flags(mocked_check_call):
```
ca-scribner, line 79: Should we add a similar test for upgrade as well?  Or is that overkill?

I know upgrade gets its istio install flags from the same property as install does, so atm it feels well tested.  But that feels like we get our test coverage by knowing an implementation detail that isn't guaranteed.  wdyt?  If it feels like overkill, I'm fine ignoring

charms/istio-pilot/src/istioctl.py
DnPlas: Oh, I kept alphabetical order here, but sure, I will change it.

charms/istio-pilot/tests/unit/test_istioctl.py
DnPlas: Yeah, I had the same feeling as you, but decided not to add one exactly because of the things you are mentioning. In the end we will be testing that the expected args are passed, which is already tested in this test case.

charms/istio-pilot/src/istioctl.py
DnPlas: Yeah, I can definitely do that.

charms/istio-pilot/tests/unit/test_istioctl.py
ca-scribner: Yeah fine with me!

charms/istio-pilot/src/istioctl.py
```diff
@@ -42,7 +44,8 @@ def _istioctl_flags(self):
             "--set",
             f"values.global.istioNamespace={self._namespace}",
         ]
-        istioctl_flags.extend(self._istioctl_extra_flags)
+        if self._istioctl_extra_flags:
+            istioctl_flags.extend(self._istioctl_extra_flags)
```
ca-scribner, line 48: Nothing against this, but imo it isn't needed since the `init` sets this private variable.  But if you want it, keep it!

ca-scribner (approved):
lgtm, apart from the CI failures which I think are unrelated and will be fixed in another PR.  Once those are working and merged in, this should be good to go

charms/istio-pilot/src/istioctl.py
DnPlas: Oh, this was a change I did in the interim, I'll revert it back to how it was. Good catch!

ca-scribner (approved):
lgtm!

PR #364, opened 2024-01-03
ci: use charmcraft latest/candidate instead of latest/edge

DnPlas:
Use the latest candidate version of charmcraft to avoid potential issues with the edge version of charmcraft.
This commit is specifically reacting to canonical/craft-application#167

> NOTE: this change makes sure that the repository is aligned with our other repositories, most of them (with very few exceptions) point to latest/candidate.

### DIFF
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -63,7 +63,7 @@ jobs:
           channel: ${{ matrix.microk8s-versions }}
           microk8s-addons: "dns hostpath-storage rbac metallb:10.64.140.43-10.64.140.49"
           juju-channel: 3.1/stable
-          charmcraft-channel: latest/edge
+          charmcraft-channel: latest/candidate
   
       - name: Run integration tests
         run: |
@@ -130,7 +130,7 @@ jobs:
           channel: ${{ matrix.microk8s-versions }}
           microk8s-addons: "dns hostpath-storage rbac metallb:10.64.140.43-10.64.140.49"
           juju-channel: 3.1/stable
-          charmcraft-channel: latest/edge
+          charmcraft-channel: latest/candidate
  
       - name: Run observability integration tests
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/integrate.yaml
```diff
@@ -63,7 +63,7 @@ jobs:
           channel: ${{ matrix.microk8s-versions }}
           microk8s-addons: "dns hostpath-storage rbac metallb:10.64.140.43-10.64.140.49"
           juju-channel: 3.1/stable
-          charmcraft-channel: latest/edge
+          charmcraft-channel: latest/candidate
```
misohu, line 66: Maybe we can link the issue in the comment

.github/workflows/integrate.yaml
```diff
@@ -130,7 +130,7 @@ jobs:
...
```
misohu, line 133: Maybe we can link the issue in the comment

.github/workflows/integrate.yaml
ca-scribner: I like putting this to a more stable version, but I don't like that we're doing it in one repo.  This should be across all repos, or have a linked issue like @misohu mentioned saying why it is different from the others and when we'll switch it back.

DnPlas: In other repositories we have pinned it to `latest/candidate`, though. I think this repo is one of the few (if not the only one) where we use edge.

misohu: I was just wondering like If I am first time reading this code I would be wondering why is this candidate why not stable :)

DnPlas: @misohu gotcha, a while ago the team agreed to pin the `latest/candidate` to receive the latest changes in an "almost stable" channel without having to wait for a full release cycle, preventing our CI to break. The change in this repository is to align to most of our other repositories, so if we decide to go with something other than latest/candidate, we can start the conversation.

ca-scribner: Resolving this because we talked offline and are in agreement since our other repos use latest/candidate.

.github/workflows/integrate.yaml
ca-scribner: Resolving as per [this thread](https://github.com/canonical/istio-operators/pull/364#discussion_r1440366926)

ca-scribner (commented):
lgtm in principle, but the CI is failing when using charmcraft so something might not be quite right yet

ca-scribner (approved):
lgtm!

PR #365, opened 2024-01-04
feat: enable Istio CNI plugin by default

### LINKED ISSUE #356: Provide a generic solution for the Istio CNI plugin init-container limitation
### What needs to get done

Because of the Istio CNI plugin [limitations](https://istio.io/latest/docs/setup/additional-setup/cni/#compatibility-with-application-init-containers), some workloads may never start as they may have init-containers that require network connectivity before the Istio sidecar proxy container is ready. 

The goal of this task is to provide a solution that automatically injects annotations to the `Pods` in the Istio mesh to exclude the redirection of certain IP ranges. This solution has been tested in #354 and it can be extended to all workloads in a `Namespace`.

## Proposed solution

Have a webhook that automatically add the required annotations.

### Why it needs to get done

To provide a solution to the limitations that we will be introducing by installing the Istio CNI plugin by default in the Charmed Kubeflow bundle.

DnPlas:
feat: enable Istio CNI plugin by default
    
This commit enables the Istio CNI plugin by default in all new
deployments of istio-pilot provided the required charm configuration is
present; otherwise the control plane remains intact.
It is also ensured that the required configurations are present when
upgrading to future versions so the plugin is correctly installed on
existing control planes (versions <1.17).
    
Fixes #356
Part of #351

#### Testing instructions

Env:
* microk8s 1.25-strict/stable
* microk8s enable dns hostpath-storage rbac metallb:10.64.140.43-10.64.140.49
* juju 3.1
* juju controller same version ^

<details>
<summary> Test charm installation </summary>

1. Set env variables
```
DEFAULT_GATEWAY="test-gateway"
KNATIVE_CHANNEL="latest/edge"
```
2. Build and deploy `istio-pilot` from this branch ` juju deploy ./istio-pilot_ubuntu-20.04-amd64.charm --config default-gateway=${DEFAULT_GATEWAY} --trust --config cni-bin-dir=<path to cni bin dir> --config cni-conf-dir=<path to cni bin dir>`

For microk8s these values are:
* /var/snap/microk8s/current/opt/cni/bin
* /var/snap/microk8s/current/args/cni-network

3. Deploy `istio-gateway` from Charmhub `juju deploy istio-gateway istio-ingressgateway --config kind="ingress" --channel latest/edge --trust`
4. Relate istio apps ` juju relate istio-pilot istio-ingressgateway`
5. Wait for them to be active and idle
6. Check that a new Pod (istio-cni-xxxx) in the model namespace is created.

##### Test integration with other charms and fix for #356 
1. Deploy kserve, knative and profile operators
```
juju deploy knative-operator --channel ${KNATIVE_CHANNEL} --trust
juju deploy knative-serving --config namespace="knative-serving" --config istio.gateway.namespace=${MODEL_NAME} --config istio.gateway.name=${DEFAULT_GATEWAY} --channel ${KNATIVE_CHANNEL} --trust
juju deploy kubeflow-profiles --channel latest/edge --trust
juju deploy kserve-controller --channel latest/edge --config deployment-mode="serverless" --trust
```
2. Add required relations
```
juju relate istio-pilot:gateway-info kserve-controller:ingress-gateway
juju relate kserve-controller:local-gateway knative-serving:local-gateway
```
3. Create a profile `kubectl apply -f https://raw.githubusercontent.com/canonical/kubeflow-profiles-operator/main/tests/integration/profile.yaml`
4. Create a `ISVC` in the recently created profile namespace `kubectl -f https://raw.githubusercontent.com/canonical/kserve-operators/main/charms/kserve-controller/tests/integration/sklearn-iris.yaml -nprofilename`
5. Check the pod for the `isvc` goes to a `Running` state and the `isvc` readiness is True. You could also perform inference following [this](https://raw.githubusercontent.com/canonical/kserve-operators/main/charms/kserve-controller/tests/integration/sklearn-iris.yaml) guide.
</details>

<details>
<summary> Test charm upgrade </summary>

1. Set env variables
```
DEFAULT_GATEWAY="test-gateway"
```

2. Deploy `istio-pilot` from Charmhub `juju deploy istio-pilot --channel latest/edge --trust`

3. Deploy `istio-gateway` from Charmhub `juju deploy istio-gateway istio-ingressgateway --config kind="ingress" --channel latest/edge --trust`

4. Relate istio apps ` juju relate istio-pilot istio-ingressgateway`

5. Wait for them to be active and idle

7. Build and refresh `istio-pilot` from this branch ` juju refresh istio-pilot  --path=./istio-pilot_ubuntu-20.04-amd64.charm`

8. The upgrade by itself won't install the cni-plugin, we must provide the cni configurations for the plugin to be enabled:

```
juju config istio-pilot cni-bin-dir=<path to cni bin dir> 
juju config istio-pilot cni-conf-dir=<path to cni bin dir>`
```

For microk8s these values are:
* /var/snap/microk8s/current/opt/cni/bin
* /var/snap/microk8s/current/args/cni-network

9. To verify the plugin was installed, we can check the pods in the model namespace, there should be an `istio-cni-xxxx` Pod present. This is not the case when the plugin is disabled. To do a more extensive test, you can follow the steps in "Test integration with other charms and fix for #356" from a previos section.

</details>
TODO:

- [x] Fix unit tests and add test cases where needed
- [x] Add manual testing instructions

### DIFF
--- a/charms/istio-pilot/config.yaml
+++ b/charms/istio-pilot/config.yaml
@@ -1,4 +1,12 @@
 options:
+  cni-bin-dir:
+    type: string
+    default: ''
+    description: Path to CNI binaries, e.g. /opt/cni/bin. This configuration option will only work when first deploying istio.
+  cni-conf-dir:
+    type: string
+    default: ''
+    description: Path to conflist files describing the CNI configuration, e.g. /etc/cni/net.d. This configuration option will only work when first deploying istio.
   default-gateway:
     type: string
     default: istio-gateway
--- a/charms/istio-pilot/config.yaml
+++ b/charms/istio-pilot/config.yaml
@@ -2,11 +2,17 @@ options:
   cni-bin-dir:
     type: string
     default: ''
-    description: Path to CNI binaries, e.g. /opt/cni/bin. This configuration option will only work when first deploying istio.
+    description: >
+      Path to CNI binaries, e.g. /opt/cni/bin. If not provided, the Istio control plane will be installed/upgraded with the Istio CNI plugin disabled.
+      This path depends on the Kubernetes installation, please refer to https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/
+      for information to find out the correct path.
   cni-conf-dir:
     type: string
     default: ''
-    description: Path to conflist files describing the CNI configuration, e.g. /etc/cni/net.d. This configuration option will only work when first deploying istio.
+    description: Path to conflist files describing the CNI configuration, e.g. /etc/cni/net.d. If not provided, the Istio control plane will be installed/upgraded
+      with the Istio CNI plugin disabled.
+      This path depends on the Kubernetes installation, please refer to https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/
+      for information to find out the correct path.
   default-gateway:
     type: string
     default: istio-gateway
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -110,10 +110,11 @@ class Operator(CharmBase):
         self._field_manager = "lightkube"
 
         # Instantiate a CertHandler
+        self.peer_relation_name = "peers"
         self._cert_handler = CertHandler(
             self,
             key="istio-cert",
-            peer_relation_name="peers",
+            peer_relation_name=self.peer_relation_name,
             cert_subject=self._cert_subject,
         )
 
@@ -296,9 +297,11 @@ class Operator(CharmBase):
         handled_errors = []
 
         # Call upgrade_charm in case there are new configurations that affect the control plane
+        # only if the CNI configurations have been provided and have changed from a previous state
         # This is useful when there is a missing configuration during the install process
         try:
-            self.upgrade_charm(event)
+            if self._cni_config_changed:
+                self.upgrade_charm(event)
         except GenericCharmRuntimeError as err:
             handled_errors.append(err)
 
@@ -871,6 +874,44 @@ class Operator(CharmBase):
     def _check_cni_configurations(self) -> bool:
         """Return True if the necessary CNI configuration options are set, False otherwise."""
         if self.model.config["cni-conf-dir"] and self.model.config["cni-bin-dir"]:
+            return True
+        return False
+
+    def _cni_config_changed(self):
+        """
+        Returns True if any of the CNI configuration options has changed from a previous state,
+        False otherwise.
+        """
+        # The peer relation is required to store values, if it does not exist because it was
+        # removed by accident, the charm should fail
+        rel = self.model.get_relation(self.peer_relation_name, None)
+        if not rel:
+            raise GenericCharmRuntimeError(
+                "The istio-pilot charm requires a peer relation, make sure it exists."
+            )
+
+        # Get current values of the configuration options
+        current_cni_bin_dir = rel.data[self.unit].get("cni-bin-dir", None)
+        current_cni_conf_dir = rel.data[self.unit].get("cni-conf-dir", None)
+
+        # Update the values based on the configuration options
+        rel.data[self.unit].update({"cni-bin-dir": self.model.config["cni-bin-dir"]})
+        rel.data[self.unit].update({"cni-conf-dir": self.model.config["cni-conf-dir"]})
+
+        new_cni_bin_dir = rel.data[self.unit].get("cni-bin-dir", None)
+        new_cni_conf_dir = rel.data[self.unit].get("cni-conf-dir", None)
+
+        # Compare current vs new values and decide whether they have changed from a previous state
+        cni_bin_dir_changed = False
+        cni_conf_dir_changed = False
+        if current_cni_bin_dir != new_cni_bin_dir:
+            cni_bin_dir_changed = True
+
+        if current_cni_conf_dir != new_cni_conf_dir:
+            cni_conf_dir_changed = True
+
+        # If any of the configuration options changed, return True
+        if cni_bin_dir_changed or cni_conf_dir_changed:
             return True
         return False
 
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -90,16 +90,6 @@ RETRY_FOR_15_MINUTES = tenacity.Retrying(
     wait=tenacity.wait_fixed(2),
     reraise=True,
 )
-CNI_MISSING_WARNING = (
-    "This version of istio-pilot requires the cni-bin-dir and cni-conf-dir "
-    "configurations to be set in order to enable the Istio CNI plugin. "
-    "The installation/upgrade with the Istio CNI plugin will be skipped. "
-    "If you want to enable this feature, please provide the required configuration."
-)
-CNI_MISSING_MSG = (
-    "Installing/Upgrading Istio control plane without Istio CNI plugin. "
-    "See juju-debug logs for more details."
-)
 
 
 class Operator(CharmBase):
@@ -240,16 +230,10 @@ class Operator(CharmBase):
 
     def install(self, _):
         """Install charm."""
-        install_message = "Deploying Istio control plane with Istio CNI plugin."
 
-        # Verify the CNI configurations are set, otherwise we cannot continue with the
-        # Control Plane installation. Deferring the event is the only way we can prevent
-        # an attempt to install the Control Plane w/o these required values.
-        if not self._check_cni_configurations():
-            self.log.warning(CNI_MISSING_WARNING)
-            install_message = CNI_MISSING_MSG
-
-        self._log_and_set_status(MaintenanceStatus(install_message))
+        self._log_and_set_status(
+            MaintenanceStatus("Deploying Istio control plane with Istio CNI plugin.")
+        )
 
         # Call the istioctl wrapper to install the Istio Control Plane
         istioctl = Istioctl(
@@ -272,8 +256,9 @@ class Operator(CharmBase):
     def reconcile(self, event):
         """Reconcile the state of the charm.
 
-        This is the main entrypoint for the charm.  It:
+        This is the main entrypoint for the method.  It:
         * Checks if we are the leader, exiting early with WaitingStatus if we are not
+        * Upgrades the Istio control plane if changes were made to the CNI plugin configurations
         * Sends data to the istio-pilot relation
         * Reconciles the ingress-auth relation, establishing whether we need authentication on our
           ingress gateway
@@ -373,13 +358,7 @@ class Operator(CharmBase):
 
         Supports upgrade of exactly one minor version at a time.
         """
-        upgrade_message = "Upgrading Istio control plane."
-        # Verify the CNI configurations are set, otherwise we cannot continue with the
-        # Control Plane installation. Deferring the event is the only way we can prevent
-        # an attempt to install the Control Plane w/o these required values.
-        if not self._check_cni_configurations():
-            self.log.warning(CNI_MISSING_WARNING)
-            upgrade_message = CNI_MISSING_MSG
+        self._log_and_set_status(MaintenanceStatus("Upgrading Istio control plane."))
 
         istioctl = Istioctl(
             ISTIOCTL_PATH,
@@ -387,7 +366,6 @@ class Operator(CharmBase):
             ISTIOCTL_DEPOYMENT_PROFILE,
             istioctl_extra_flags=self._istioctl_extra_flags,
         )
-        self._log_and_set_status(MaintenanceStatus(upgrade_message))
 
         # Check for version compatibility for the upgrade
         try:
@@ -873,9 +851,7 @@ class Operator(CharmBase):
 
     def _check_cni_configurations(self) -> bool:
         """Return True if the necessary CNI configuration options are set, False otherwise."""
-        if self.model.config["cni-conf-dir"] and self.model.config["cni-bin-dir"]:
-            return True
-        return False
+        return self.model.config["cni-conf-dir"] and self.model.config["cni-bin-dir"]
 
     def _cni_config_changed(self):
         """
@@ -911,9 +887,7 @@ class Operator(CharmBase):
             cni_conf_dir_changed = True
 
         # If any of the configuration options changed, return True
-        if cni_bin_dir_changed or cni_conf_dir_changed:
-            return True
-        return False
+        return cni_bin_dir_changed or cni_conf_dir_changed
 
 
 def _get_gateway_address_from_svc(svc):
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -90,6 +90,16 @@ RETRY_FOR_15_MINUTES = tenacity.Retrying(
     wait=tenacity.wait_fixed(2),
     reraise=True,
 )
+CNI_MISSING_WARNING = (
+    "This version of istio-pilot requires the cni-bin-dir and cni-conf-dir "
+    "configurations to be set in order to enable the Istio CNI plugin. "
+    "The installation/upgrade with the Istio CNI plugin will be skipped. "
+    "If you want to enable this feature, please provide the required configuration."
+)
+CNI_MISSING_MSG = (
+    "Installing/Upgrading Istio control plane without Istio CNI plugin. "
+    "See juju-debug logs for more details."
+)
 
 
 class Operator(CharmBase):
@@ -192,7 +202,6 @@ class Operator(CharmBase):
 
         # Extra flags to pass to the istioctl install command
         # These flags will configure the container images used by the control plane
-        # and also configure the CNI plugin behaviour
         extra_flags = [
             "--set",
             f"values.pilot.image={pilot_image}",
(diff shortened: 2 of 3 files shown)


charms/istio-pilot/config.yaml
```diff
@@ -1,4 +1,12 @@
 options:
+  cni-bin-dir:
+    type: string
+    default: ''
+    description: Path to CNI binaries, e.g. /opt/cni/bin. This configuration option will only work when first deploying istio.
+  cni-conf-dir:
+    type: string
+    default: ''
+    description: Path to conflist files describing the CNI configuration, e.g. /etc/cni/net.d. This configuration option will only work when first deploying istio.
```
ca-scribner, line 9: Can we add a link to something that describes how to get these, or example usage?  If you intend on adding a readme or something else later, that works too

charms/istio-pilot/src/charm.py
```diff
@@ -206,12 +206,48 @@ def install(self, _):
...
+                "This version of istio-pilot requires the cni-bin-dir and cni-conf-dir "
+                "configurations to be set in order to enable the Istio CNI plugin. "
+                "The installation with the Istio CNI plugin will be skipped. "
+                "If you want to enable this feature, please provide the required configuration."
+            )
+            install_message = (
+                "Deploying Istio control plane without Istio CNI plugin. "
+                "See juju-debug logs for more details."
+            )
+
```
ca-scribner, line 242: This warning feels out of place.  I think you're trying to recommend to people they should use the CNI plugin, but it says this version of istio-pilot requires CNI, but since you haven't asked for it we'll deploy without it.  "requires" in particular feels like the wrong word since we don't really require it

charms/istio-pilot/src/charm.py
```diff
@@ -251,6 +287,13 @@ def reconcile(self, event):
         # so that we can report them at the end.
         handled_errors = []
 
+        # Call upgrade_charm in case there are new configurations that affect the control plane
+        # This is useful when there is a missing configuration during the install process
+        try:
+            self.upgrade_charm(event)
+        except GenericCharmRuntimeError as err:
+            handled_errors.append(err)
```
ca-scribner, line 295: Is there a reason why this is needed?  I don't have a specific case where I think it breaks, but I worry about it because it combines a few different things.  Telling `istioctl` it is allowed to overwrite istio on every config change, relation change, etc feels like a lot

charms/istio-pilot/src/charm.py
```diff
@@ -319,8 +362,29 @@ def upgrade_charm(self, _):
...
+            self.log.warning(
+                "This version of istio-pilot requires the cni-bin-dir and cni-conf-dir "
+                "configurations to be set in order to enable the Istio CNI plugin. "
+                "The installation with the Istio CNI plugin will be skipped. "
+                "If you want to enable this feature, please provide the required configuration."
+            )
+            upgrade_message = (
+                "Upgrading Istio control plan without Istio CNI plugin. "
+                "See `juju debug-log` for details."
+            )
```
ca-scribner, line 379: If this message is really needed, we should make it common with the message displayed in install.  They can come from the same constant or helper function.

I'm not sure this message adds much, and if we also add a call to `upgrade_charm` in reconcile ([here](https://github.com/canonical/istio-operators/pull/365/files#diff-d3a4e684bd1a3a1822df5ab7605cfce00c97cd7b52f44ed770526c4f4f80eacdR293)), the logs are going to have this warning on every config-changed, relation-joined, relation-changed, etc.    That feels like a lot

charms/istio-pilot/config.yaml
```diff
@@ -1,4 +1,12 @@
 options:
+  cni-bin-dir:
+    type: string
+    default: ''
+    description: Path to CNI binaries, e.g. /opt/cni/bin. This configuration option will only work when first deploying istio.
```
ca-scribner, line 5: If we've [added `upgrade_charm` to reconcile](https://github.com/canonical/istio-operators/pull/365/files#diff-d3a4e684bd1a3a1822df5ab7605cfce00c97cd7b52f44ed770526c4f4f80eacdR293), is this statement still true?  I'm guessing that as soon as we change this, there will be a reconcile that redeploys istio with this setting.

DnPlas: Correct, the message will be different here. I will update it to reflect what is actually happening. Good catch!

charms/istio-pilot/config.yaml
DnPlas: I think both information in the README and a link here are useful. Will update this file and I will add info in the README (see https://github.com/canonical/istio-operators/issues/366).

charms/istio-pilot/src/charm.py
```diff
@@ -206,12 +206,48 @@ def install(self, _):
...
+            extra_flags.extend(
+                [
+                    "--set",
+                    "components.cni.enabled=true",
+                    "--set",
+                    f"values.cni.cniBinDir={self.model.config['cni-bin-dir']}",
+                    "--set",
+                    f"values.cni.cniConfDir={self.model.config['cni-conf-dir']}",
+                    "--set",
+                    "values.sidecarInjectorWebhook.injectedAnnotations.traffic\.sidecar\.istio\.io/excludeOutboundIPRanges=0.0.0.0/0",  # noqa
```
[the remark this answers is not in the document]
DnPlas: That makes sense, will add both a comment in the code and in this PR. In the issue I believe it is well documented.

charms/istio-pilot/src/charm.py
DnPlas: The reason we now have to call upgrade in the reconcile function is because there might be cases where users provide none or only one of the two required cni configurations during install. In that case, we don't want to install the control plane with the cni plugin enabled without these configurations, as we know it will 100% end in error. Instead of us blocking on install and deferring the event in the absence of the required configs, we warn the user that the installation is going to be w/o the plugin and give an option to later provide these configs.

The other option is to defer the install event if these two configs are missing.

Let me know what you think, if you have an alternative, I'll be glad to discuss it.

charms/istio-pilot/src/charm.py
DnPlas: Yeah, I kind of also wanted to abstract it to a common helper or constant, I'll change it.

> I'm not sure this message adds much, and if we also add a call to upgrade_charm in reconcile ([here](https://github.com/canonical/istio-operators/pull/365/files#diff-d3a4e684bd1a3a1822df5ab7605cfce00c97cd7b52f44ed770526c4f4f80eacdR293)), the logs are going to have this warning on every config-changed, relation-joined, relation-changed, etc. That feels like a lot

I want to be very verbose about this, actually, I think it must be clear what kind of installation users are receiving and why.

charms/istio-pilot/src/charm.py
DnPlas: Actually, what I am saying is that in order to enable the CNI plugin, this charm requires the cni configs; otherwise the plugin will be disabled. Is the message not clear bout it? I can re-phrase it if you think this should be different.

charms/istio-pilot/src/charm.py
```diff
@@ -251,6 +295,13 @@ def reconcile(self, event):
         # so that we can report them at the end.
         handled_errors = []
 
+        # Call upgrade_charm in case there are new configurations that affect the control plane
+        # This is useful when there is a missing configuration during the install process
+        try:
+            self.upgrade_charm(event)
```
ca-scribner, line 301: Assuming we do keep this in the reconcile function, can we also update the reconcile docstring to mention istio will also be reconciled?

charms/istio-pilot/src/charm.py
ca-scribner: Ah ok, then do we need these messages at all?  If this is an option that the user can enable, should the charm keep saying "you haven't enabled this option" on every reconcile?  That seems like a lot

The config and readme will say "you can enable CNI by setting these two configs, otherwise it'll use the standard version".  Isn't that enough notice to the user?  I can't think of other things in our charms we have that are optional but we keep reminding the user about like this

charms/istio-pilot/src/charm.py
ca-scribner: ok, so enable users to take their existing charm and turn on CNI, we need this check in reconcile.  Gotcha  ...  but I'm still nervous :p  

Do we know if `istioctl upgrade` is idempotent (eg: that running it many times will not cause things to restart or interrupt anything)?  I'm searching for docs mentioning this but I don't see anything.  From playing for a few minutes locally I **think** it is idempotent?  I did `istioctl install ...`, did `istioctl upgrade ...` a few times and didn't see anything bad happen.

ca-scribner: Our [upgrade guide](https://charmed-kubeflow.io/docs/upgrade-17-18#heading--upgrade--istio) tells users to:
* scale down the ingress gateway
* upgrade istio
* scale the ingress back up again
I think that's because of [this note](https://github.com/canonical/istio-operators/blob/main/charms/istio-pilot/README.md#user-content-fn-1-126dacc6f56e23c6bc138b65a6326e5b) about how `istioctl upgrade` sometimes hangs if our ingress gateway exists.  If this is true here, I don't think we can have `upgrade` called in `reconcile`.  But iirc we only saw the hanging upgrade issue when upgrading istio 1.11->1.12 so we might be ok.  I'm not sure how to robustly test it though - I don't recall if it was intermittent or it always occurred.  What do you think?

charms/istio-pilot/src/charm.py
ca-scribner, line 1: If we do leave the call to `upgrade_charm` in the `reconcile` function, do we need to observe the `upgrade-charm` event at all?  `upgrade-charm` is followed by `config-changed`, so it might be redundant.

charms/istio-pilot/src/charm.py
ca-scribner: I think this conversation and [the above thread](https://github.com/canonical/istio-operators/pull/365#discussion_r1443356485) have now merged - I'll keep replying to the other and we can apply whatever we decide there here too

charms/istio-pilot/src/charm.py
DnPlas: There is a second option, in which we could use the `istioctl install -f <overlay-file.yaml>` to update the configuration the installation. This is how they recommend doing it ([docs](https://istio.io/latest/docs/setup/additional-setup/customize-installation/)).
I'll test that option and update the PR accordingly.

DnPlas: I tried a different approach: I am still calling the `upgrade_charm` method in `reconcile`, but the condition to run it is if the configuration that has changed is actually the CNI configurations, that way we avoid running upgrades on every single reconcile. 
The reason why I still want to use `istioctl upgrade` to add these configurations, is because that is the most straightforward way to apply changes while ignoring the things already installed that don't require a change.

What do you think?

ca-scribner: I like the added check on whether the setting has changed - it adds a tiny bit of complexity but I think its safe and protects against anything bad going on in repeated `istioctl upgrade` calls.  I think you're right that firing a `istioctl upgrade` when things haven't changed is a noop, but nice to just avoid it.

ca-scribner: I think we can make `_cni_config_changed` generic and put it into chisme then import it here.  Something like (I added StoredState here too just for discussion, but it could be omitted):

```python
def has_config_changed(config_name, storage_method="storedstate"):
  if storage_method == "storedstate":
    current_value = get_it_from_stored_state
  elif storage_method == "peer_relation":
    rel = self.model.get_relation(self.peer_relation_name, None)
    if not rel:
      # If we can't confirm this is a cache hit, this is a cache miss and we return True
      return True
    # Maybe add some prefix on config_name below?
    current = rel.data[self.unit].get(config_name, None)
  new = self.model.config[config_name]
  if new == old:
    # did not change
    return False
  else:
    # did change
    # update cache
    ...
    return True
```

Then in the istio charm, you can do
```python
if has_config_changed("cni-bin-dir") or has_config_changed("cni-conf-dir"):
  do_update()
```

`has_config_changed` can live in chisme and then we can use it elsewhere too

charms/istio-pilot/src/charm.py
```diff
@@ -251,6 +296,15 @@ def reconcile(self, event):
...
 
+        # Call upgrade_charm in case there are new configurations that affect the control plane
+        # only if the CNI configurations have been provided and have changed from a previous state
+        # This is useful when there is a missing configuration during the install process
+        try:
+            if self._cni_config_changed:
+                self.upgrade_charm(event)
+        except GenericCharmRuntimeError as err:
+            handled_errors.append(err)
+
```
ca-scribner, line 307: Since this can only trigger during a config change, can we move it to its own handler?  That way the logic for reconciling  the relations is separate from the reconfiguring of the istio daemon?  So something like

```python

class Operator(CharmBase):
  def __init__(self, *args):
    ...
    # Handle when istio deployment configuration is changed
    self.framework.observe(self.on.config_changed, self.update_istio_cni_configurations)

  def update_istio_cni_configurations(self, _):
    try:
      if self._cni_config_changed:
        self.upgrade_charm(event)
    except GenericCharmRuntimeError as err:
      handled_errors.append(err)
```

This'll help keep logic as encapsulated as we can

charms/istio-pilot/src/charm.py
```diff
@@ -804,6 +871,50 @@ def _log_and_set_status(self, status):
...
+        if current_cni_bin_dir != new_cni_bin_dir:
+            cni_bin_dir_changed = True
+
+        if current_cni_conf_dir != new_cni_conf_dir:
+            cni_conf_dir_changed = True
+
+        # If any of the configuration options changed, return True
+        if cni_bin_dir_changed or cni_conf_dir_changed:
+            return True
+        return False
```
ca-scribner, line 916: tiny nit simplification, ignore if you want
```suggestion
        return cni_bin_dir_changed or cni_conf_dir_changed
```

charms/istio-pilot/src/charm.py
DnPlas: Yeah, the more I think of this the more I agree. I think it is fair to assume people will know and they are informed about the type of installation they are receiving. Will remove in both install and upgrade methods.

charms/istio-pilot/src/charm.py
DnPlas: I agree this can be a reusable thing, but I prefer to do that refactor in a follow up PR because it will mean the introduction of unrelated changes to this istio cni install effort (things like requirements updates, potentially removal of a unit tests). If you don't mind, I'll do it after we merge this :)

charms/istio-pilot/src/charm.py
DnPlas: Would that be possible though? We are already [observing config changed](https://github.com/canonical/istio-operators/blob/e574d9ecb2578b81e58338a521e2ad3ec6412345/charms/istio-pilot/src/charm.py#L133) events and tying them to the reconcile method.

charms/istio-pilot/src/charm.py
DnPlas: True, thanks!

charms/istio-pilot/src/charm.py
```diff
@@ -804,6 +871,50 @@ def _log_and_set_status(self, status):
...
+        Returns True if any of the CNI configuration options has changed from a previous state,
+        False otherwise.
+        """
+        # The peer relation is required to store values, if it does not exist because it was
+        # removed by accident, the charm should fail
+        rel = self.model.get_relation(self.peer_relation_name, None)
+        if not rel:
+            raise GenericCharmRuntimeError(
+                "The istio-pilot charm requires a peer relation, make sure it exists."
+            )
```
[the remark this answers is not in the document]
DnPlas: First of all, this check is 99.99% ensuring that we don't remove the peer relation by mistake from the charm's metadata, as users usually won't be able to manipulate this type of relation.

I agree that perhaps raising an exception and halting the charm may put the charm in an unrecoverable state, which is not ideal. I wonder if this check should be elsewhere.

Your suggestion could work, but I worry that we will then be upgrading on every single reconcile, which is something we do not want. This issue can be solved by returning False, but then returning False will also prevent the upgrade from happening if there was in fact a config change in those cni values.

charms/istio-pilot/src/charm.py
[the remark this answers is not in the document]
DnPlas: Documented.

charms/istio-pilot/config.yaml
DnPlas: Updated.

charms/istio-pilot/src/charm.py
DnPlas: There is no guarantee that the upgrade charm method is going to be called in the reconcile function because I have added a check that only runs it if there was a config changed in CNI configs. I think we should keep it.

charms/istio-pilot/src/charm.py
ca-scribner: I'd have preferred it here (I don't like adding something we immediately plan to remove), but that's ok.  We can resolve this

charms/istio-pilot/src/charm.py
ca-scribner: Yup with that new check, this doesn't apply anymore

ca-scribner: Yup with that new check, this doesn't apply anymore

charms/istio-pilot/src/charm.py
ca-scribner: Definitely possible.  For example, the observability libs we use subscribe to the same events as some of our own handlers.  Juju just runs everything that subscribes in sequence (not sure if the order is defined)

But, given that we just talked today about splitting this charm into two (istio deployment and istio integrator), I say we just ignore this comment for now as soon it wont matter. I'll resolve this conversation

charms/istio-pilot/src/charm.py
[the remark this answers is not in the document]
ca-scribner: I think the base question we ask here is:
* if config has not changed, do nothing
* else: upgrade

If we can confirm that the config has not changed, we skip.  Otherwise, we upgrade.  So that's why I see both `old!=new` and `cache does not exist` as the same event - I don't see a need to distinguish between a confirmed config change and an unknown config change.

However, given that we are discussing splitting this charm into two (an istio charm and an istio integrator charm), I'm fine with either way as I think we'll probably change it soon.

charms/istio-pilot/src/charm.py
ca-scribner: (but lets make an issue, maybe in chisme, saying this could be turned general and maybe linking this thread)

ca-scribner (approved):
lgtm!  great job @DnPlas

PR #371, opened 2024-01-25
feat: set meshConfig.accessLogFile configuration for exposing logs

### LINKED ISSUE #370: Envoy access logs are not exposed by default
### Bug Description

Currently all the envoy sidecars don't expose [access logs](https://istio.io/latest/docs/tasks/observability/logs/access-log/)

This means that if a sidecar drops a request (i.e. because of an AuthorizationPolicy) then we'll never see logs in the `istio-proxy` sidecar.

This makes it super difficult to debug

### To Reproduce

1. Create a second profile, for the same user
2. Create `notebook-a` in the first profile
3. Create `notebook-b` in the second profile
4. Make a curl request from `notebook-b` to the K8s SVC of `notebook-a`
5. Check the logs of the `istio-proxy` sidecar of `notebook-a`

At this point we should see a log like this, but we DONT
```
[2024-01-17T23:31:26.988Z] "GET / HTTP/1.1" 403 - rbac_access_denied_matched_policy[none] - "-" 0 19 0 - "-" "curl/7.81.0" "69a1c216-884c-40c2-b842-50a6d6f92509" "opensearch.admin.svc.cluster.local" "-" inbound|8888|| - 10.1.252.3:8888 10.1.252.4:41438 outbound_.80_._.opensearch.admin.svc.cluster.local -
```

### Environment

Charmed Kubeflow 1.8
Istio Pilot 1.7

### Relevant Log Output

```shell
2024-01-17T23:31:04.945480Z     info    Envoy proxy is ready
```

### Additional Context

This can be resolved by setting `--set meshConfig.accessLogFile=/dev/stdout` when installing Istio
https://istio.io/latest/docs/tasks/observability/logs/access-log/#using-mesh-config

cc @ca-scribner @NohaIhab

DnPlas:
This setting allow users to access debug logs.

Fixes #370

#### Manual testing

As described in #370's description.

### DIFF
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -193,6 +193,7 @@ class Operator(CharmBase):
 
         # Extra flags to pass to the istioctl install command
         # These flags will configure the container images used by the control plane
+        # As well as set the access log files for help during debugging
         extra_flags = [
             "--set",
             f"values.pilot.image={pilot_image}",
@@ -204,6 +205,8 @@ class Operator(CharmBase):
             f"values.global.proxy.image={global_proxy_image}",
             "--set",
             f"values.global.proxy_init.image={global_proxy_init_image}",
+            "--set",
+            "meshConfig.accessLogFile=/dev/stdout",
         ]
 
         # The following are a set of flags that configure the CNI behaviour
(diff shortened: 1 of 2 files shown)


orfeas-k:
Is this change expected to take place if I `juju refresh` an older revision of istio charms?

orfeas-k (approved):
LGTM! I tried in a deployment with older istio and worked there too.

DnPlas:
> Is this change expected to take place if I `juju refresh` an older revision of istio charms?

Yes, the `istioctl upgrade` command should catch the extra flags, so any version before this change should automatically have it once the `istio-pilot` charm is updated with `juju refresh`.

PR #381, opened 2024-02-20
feat: add `csr-domain-name` config option 

### LINKED ISSUE #379: Integration with charms for Let's encrypt certificates
### Bug Description

I deployed Kubeflow 1.8 and tryed to integrate Istio-pilot charm with [httpreq acme operator](https://charmhub.io/httprequest-lego-k8s) by following their how-to-guide. Unfortunately, after adding the relation, the certificate is not being created, because Let's encrypt doesn't allow provisioning the certificates for IP addresses. And the istio-pilot charm is supplying the IP address with the CSR to the certificate charm. This can be seen in the juju debug-log bellow.

It makes sense to me that istio-pilot charm is using the IP because it is not aware of the url that needs to be used in our case. Thi url is supplied only as a parameter in oidc-gatekeeper and dex-auth charms in my bundle. The same way, in kubernetes, this service is somehow no aware of this url. 
`
istio-ingressgateway                              ClusterIP      10.152.183.184   <none>                             65535/TCP                               10d
istio-ingressgateway-endpoints                    ClusterIP      None             <none>                             <none>                                  10d
istio-ingressgateway-workload                     LoadBalancer   10.152.183.89    <IP>                       80:30481/TCP,443:30282/TCP              10d
istio-pilot                                       ClusterIP      10.152.183.57    <none>                             65535/TCP                               10d
`

I'm not sure if this is a bug or is there a way to configure istio-pilot to use the url we need

### To Reproduce

1. deploy kubeflow 1.8/stable
2. juju deploy httprequest-lego-k8s
3. juju config httprequest-lego-k8s \
  server=<your ACME server> \
  email=<your email address> \
  httpreq_endpoint=<your HTTP Request endpoint> 
4. juju integrate istio-pilot httprequest-lego-k8s

### Environment

Kubeflow 1.8/stable https://github.com/canonical/bundle-kubeflow/tree/main/releases/1.8/stable/kubeflow 
Charmed Kubernetes 1.28 on Charmed Openstack Yoga
juju 3.1

### Relevant Log Output

```shell
unit-istio-pilot-0: 08:33:38 INFO juju.worker.uniter.operation ran "certificates-relation-changed" hook (via hook dispatching script: dispatch)
unit-httprequest-lego-k8s-0: 08:33:37 INFO unit.httprequest-lego-k8s/0.juju-log certificates:172: Received Certificate Creation Request for domain <IP>
unit-httprequest-lego-k8s-0: 08:33:38 ERROR unit.httprequest-lego-k8s/0.juju-log certificates:172: Exited with code 1. Stderr:
unit-httprequest-lego-k8s-0: 08:33:38 ERROR unit.httprequest-lego-k8s/0.juju-log certificates:172:     2024/02/08 08:33:37 No key found for account <customer-email>. Generating a P256 key.
unit-httprequest-lego-k8s-0: 08:33:38 ERROR unit.httprequest-lego-k8s/0.juju-log certificates:172:     2024/02/08 08:33:37 Saved key to /tmp/.lego/accounts/acme-v02.api.letsencrypt.org/<customer-email>/keys/<customer-email>.key
unit-httprequest-lego-k8s-0: 08:33:38 ERROR unit.httprequest-lego-k8s/0.juju-log certificates:172:     2024/02/08 08:33:38 [INFO] acme: Registering account for <customer-email>
unit-httprequest-lego-k8s-0: 08:33:38 ERROR unit.httprequest-lego-k8s/0.juju-log certificates:172:     2024/02/08 08:33:38 [INFO] [<IP>, istio-pilot-0.istio-pilot-endpoints.kubeflow.svc.cluster.local] acme: Obtaining bundled SAN certificate given a CSR
unit-httprequest-lego-k8s-0: 08:33:38 ERROR unit.httprequest-lego-k8s/0.juju-log certificates:172:     2024/02/08 08:33:38 Could not obtain certificates:
unit-httprequest-lego-k8s-0: 08:33:38 ERROR unit.httprequest-lego-k8s/0.juju-log certificates:172:     	acme: error: 400 :: POST :: https://acme-v02.api.letsencrypt.org/acme/new-order :: urn:ietf:params:acme:error:unsupportedIdentifier :: NewOrder request included invalid non-DNS type identifier: type "ip", value "<IP>"
```

### Additional Context

_No response_

DnPlas:
feat: enable csr-domain-name config option so istio-pilot can use it on CSRs

The istio-pilot charm already has a mechanism in place to discover the ingress gateway address from the `Service`, but it is limited to only returning IP addresses, which not all TLS certificate providers accept as a valid cert subject. Having the domain-name config option will allow users to specify the domain name they'd like to use when integrating with TLS certificate operators. This feature expands the support for integrating with TLS certificate providers that cannot issue signed certificates on a CSR that only contains an IP address (like we used to do).
This commit also adds some test coverage to test the recently added code.

Fixes #379

### DIFF
--- a/charms/istio-pilot/config.yaml
+++ b/charms/istio-pilot/config.yaml
@@ -17,6 +17,11 @@ options:
     type: string
     default: istio-gateway
     description: Name to use as a default gateway
+  domain-name:
+    default: ''
+    type: string
+    description: |
+      The domain name to be used by the charm to send a Certificate Signing Request (CSR) to a TLS certificate provider. In the absence of this configuration option, the charm will try to use the ingress gateway service hostname (if configured by a LB) or its IP address.
   image-configuration:
     default: |
       pilot-image: 'pilot'  # values.pilot.image
--- a/charms/istio-pilot/config.yaml
+++ b/charms/istio-pilot/config.yaml
@@ -13,6 +13,11 @@ options:
       with the Istio CNI plugin disabled.
       This path depends on the Kubernetes installation, please refer to https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/
       for information to find out the correct path.
+  csr-domain-name:
+    default: ''
+    type: string
+    description: |
+      The domain name to be used by the charm to send a Certificate Signing Request (CSR) to a TLS certificate provider. In the absence of this configuration option, the charm will try to use the ingress gateway service hostname (if configured by a LB) or its IP address.
   default-gateway:
     type: string
     default: istio-gateway
--- a/charms/istio-pilot/config.yaml
+++ b/charms/istio-pilot/config.yaml
@@ -17,11 +17,6 @@ options:
     type: string
     default: istio-gateway
     description: Name to use as a default gateway
-  domain-name:
-    default: ''
-    type: string
-    description: |
-      The domain name to be used by the charm to send a Certificate Signing Request (CSR) to a TLS certificate provider. In the absence of this configuration option, the charm will try to use the ingress gateway service hostname (if configured by a LB) or its IP address.
   image-configuration:
     default: |
       pilot-image: 'pilot'  # values.pilot.image
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -1,7 +1,10 @@
 #!/usr/bin/env python3
 
 import base64
+import contextlib
+import ipaddress
 import logging
+import socket
 from typing import List, Optional
 
 import tenacity
@@ -441,16 +444,38 @@ class Operator(CharmBase):
             raise ErrorWithStatus("Waiting for leadership", WaitingStatus)
 
     @property
-    def _cert_subject(self):
-        """Return the domain to be used in the CSR."""
+    def _ingress_gateway_host(self) -> Optional[str]:
+        """Return a domain name if it's configured, the ingress gateway svc address otherwise."""
+        # If the configuration option is set, we use it
+        if domain_name := self.model.config["domain-name"]:
+            return domain_name
+
+        # We try getting the ingress gateway IP address in the absence of domain name
         try:
             svc = self._get_gateway_service()
         except ApiError:
-            self.log.info("Could not retrieve the gateway service for configuring the CSR.")
+            self.log.info("Could not retrieve the gateway service for using in the CSR.")
             return None
-        gateway_address = _get_gateway_address_from_svc(svc)
-        if gateway_address:
-            return gateway_address
+        return _get_gateway_address_from_svc(svc)
+
+    @property
+    def _cert_subject(self) -> Optional[str]:
+        """Return the certificate subject to be used in the CSR."""
+        target = self._ingress_gateway_host
+
+        # Check if the target is a hostname and return if it is
+        if _is_hostname(target):
+            assert isinstance(target, str), target
+            return target
+
+        # If target is an IP address, try to get the hostname and return it
+        with contextlib.suppress(OSError, TypeError):
+            name, _, _ = socket.gethostbyaddr(target)
+            if _is_hostname(name) and not name.endswith(".svc.cluster.local"):
+                return name
+
+        # If none of the above return, the cert subject will be the IP address
+        return target if target else None
 
     @property
     def _gateway_port(self):
@@ -1055,6 +1080,17 @@ def _xor(a, b):
         return True
     else:
         return False
+
+
+def _is_hostname(value: Optional[str]) -> bool:
+    """Return True if the given value is a valid hostname, False otherwise or if it's None."""
+    if value is None:
+        return False
+    try:
+        ipaddress.ip_address(value)
+        return False
+    except ValueError:
+        return bool(value)
 
 
 if __name__ == "__main__":
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -1,10 +1,7 @@
 #!/usr/bin/env python3
 
 import base64
-import contextlib
-import ipaddress
 import logging
-import socket
 from typing import List, Optional
 
 import tenacity
@@ -447,38 +444,26 @@ class Operator(CharmBase):
     def _cert_subject(self) -> Optional[str]:
         """Return the certificate subject to be used in the CSR.
 
-        This property takes advantage of the automatic address discovery of the
-        ingress gateway address and uses it to return a name that can be used as
-        a certificate subject when integrated with a TLS certificate provider.
-        The logic is as follows:
-        * Return None if the ingress gateway service address cannot be retrieved
-        * Return a hostname if the address is not an IP (e.g. user-example.com)
-        * Return a hostname if the address is an IP, but there is a hostname that resolves to it
-        * Return the IP address if none of the above happens and the address exists
-        * Return None in any other case
+        If the csr-domain-name configuration option is set, this value is used;
+        otherwise, use the IP address or hostname of the actual ingress gateway service.
+        Lastly, if for any reason the service address cannot be retrieved, None will be returned.
         """
+        # Prioritise the csr-domain-name config option
+        if csr_domain_name := self.model.config["csr-domain-name"]:
+            return csr_domain_name
+
         # Get the ingress gateway service address
         try:
             svc = self._get_gateway_service()
         except ApiError:
-            self.log.info("Could not retrieve the gateway service for using in the CSR.")
+            self.log.info("Could not retrieve the gateway service address for using in the CSR.")
             return None
 
-        target = _get_gateway_address_from_svc(svc)
+        svc_address = _get_gateway_address_from_svc(svc)
 
-        # Check if the address is an IP, return if it is a hostname
-        if _is_not_ipaddress(target):
-            assert isinstance(target, str), target
-            return target
-
-        # If the address is an IP, try to get the hostname and return it
-        with contextlib.suppress(OSError, TypeError):
-            name, _, _ = socket.gethostbyaddr(target)
-            if _is_not_ipaddress(name) and not name.endswith(".svc.cluster.local"):
-                return name
-
-        # If none of the above return, the cert subject will be the IP address
-        return target if target else None
+        # NOTE: returning None here means that the CSR will have the unit name as cert subject
+        # this is an implementation detail of the cert_hanlder library.
+        return svc_address if svc_address else None
 
     @property
     def _gateway_port(self):
@@ -1085,17 +1070,6 @@ def _xor(a, b):
         return True
     else:
         return False
-
-
-def _is_not_ipaddress(value: Optional[str]) -> bool:
-    """Return True if the given value is not an IP address, False otherwise or if it's None."""
-    if value is None:
-        return False
-    try:
-        ipaddress.ip_address(value)
-        return False
-    except ValueError:
-        return True
 
 
 if __name__ == "__main__":
--- a/charms/istio-pilot/src/charm.py
+++ b/charms/istio-pilot/src/charm.py
@@ -444,34 +444,37 @@ class Operator(CharmBase):
             raise ErrorWithStatus("Waiting for leadership", WaitingStatus)
 
     @property
-    def _ingress_gateway_host(self) -> Optional[str]:
-        """Return a domain name if it's configured, the ingress gateway svc address otherwise."""
-        # If the configuration option is set, we use it
-        if domain_name := self.model.config["domain-name"]:
-            return domain_name
+    def _cert_subject(self) -> Optional[str]:
+        """Return the certificate subject to be used in the CSR.
 
-        # We try getting the ingress gateway IP address in the absence of domain name
+        This property takes advantage of the automatic address discovery of the
+        ingress gateway address and uses it to return a name that can be used as
+        a certificate subject when integrated with a TLS certificate provider.
+        The logic is as follows:
+        * Return None if the ingress gateway service address cannot be retrieved
+        * Return a hostname if the address is not an IP (e.g. user-example.com)
+        * Return a hostname if the address is an IP, but there is a hostname that resolves to it
+        * Return the IP address if none of the above happens and the address exists
+        * Return None in any other case
+        """
+        # Get the ingress gateway service address
         try:
             svc = self._get_gateway_service()
         except ApiError:
             self.log.info("Could not retrieve the gateway service for using in the CSR.")
             return None
-        return _get_gateway_address_from_svc(svc)
 
-    @property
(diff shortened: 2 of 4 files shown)


tests/test_bundle.py
```diff
@@ -100,7 +100,7 @@ async def test_ingress_relation(ops_test: OpsTest):
     TODO (https://github.com/canonical/istio-operators/issues/259): Change this from using a
      specific charm that implements ingress's requirer interface to a generic charm
     """
-    await ops_test.model.deploy(KUBEFLOW_VOLUMES, channel="latest/edge")
```
ca-scribner, line 103: Is this pin needed?  I'm not against it exactly, but usually our `main` CI pins against `latest`.  If there's a reason why `latest` doesn't work, we should add a comment explaining why and when we can revert back to `latest`

charms/istio-pilot/src/charm.py
```diff
@@ -441,16 +444,38 @@ def _check_leader(self):
...
+            return target
+
+        # If target is an IP address, try to get the hostname and return it
+        with contextlib.suppress(OSError, TypeError):
+            name, _, _ = socket.gethostbyaddr(target)
+            if _is_hostname(name) and not name.endswith(".svc.cluster.local"):
+                return name
+
+        # If none of the above return, the cert subject will be the IP address
+        return target if target else None
```
orfeas-k, line 478: the implementation here makes sense, but it feels a bit too manual. I can't think of a way to improve it though, that's why I 'm thinking that maybe it's as good as this gets.

orfeas-k: Should we have unit tests about `cert_subject` property? I see we 're testing the sub pieces which is good.

orfeas-k (commented):
Thank you @dnplas. I'm adding what we discussed in Matrix:
* Have you tested this? Should a reviewer test this too? If yes, how?
* Do we need to update docs for this and how it's supposed to be used?

In general the changes make sense to me and this is some good work. I have some thoughts on the `cert_subject` property, but let's see what you think as well.

DnPlas:
I expect the CI to fail until we merge #382

charms/istio-pilot/src/charm.py
```diff
@@ -441,16 +444,38 @@ def _check_leader(self):
...
+        return _get_gateway_address_from_svc(svc)
+
+    @property
+    def _cert_subject(self) -> Optional[str]:
+        """Return the certificate subject to be used in the CSR."""
+        target = self._ingress_gateway_host
+
+        # Check if the target is a hostname and return if it is
+        if _is_hostname(target):
+            assert isinstance(target, str), target
```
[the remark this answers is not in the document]
ca-scribner: If this test is meant to be here, we should change it to an if statement that throws an error instead of an `assert`.  depending on how python is executed, assert statements may be removed at runtime (see `python --help`'s `-O`)

charms/istio-pilot/src/charm.py
```diff
@@ -1057,5 +1082,16 @@ def _xor(a, b):
...
 
+def _is_hostname(value: Optional[str]) -> bool:
+    """Return True if the given value is a valid hostname, False otherwise or if it's None."""
+    if value is None:
+        return False
+    try:
+        ipaddress.ip_address(value)
+        return False
+    except ValueError:
+        return bool(value)
```
ca-scribner, line 1093: I don't know if this really matters, but this test is really `is_not_ipaddress` rather than `is_hostname`.  For example, these are not valid hostnames but pass:
* `_is_hostname("*")` returns `True`
* `_is_hostname("a"*1000)` returns `True`

ca-scribner: If its important, [this](https://stackoverflow.com/a/33214423) might be a more complete solution.  If it isn't important, renaming the test to `is_not_ipaddress` (or `is_ipaddress`) would be clearer

charms/istio-pilot/src/charm.py
```diff
@@ -441,16 +444,38 @@ def _check_leader(self):
...
+        """Return the certificate subject to be used in the CSR."""
+        target = self._ingress_gateway_host
+
+        # Check if the target is a hostname and return if it is
+        if _is_hostname(target):
+            assert isinstance(target, str), target
+            return target
+
+        # If target is an IP address, try to get the hostname and return it
+        with contextlib.suppress(OSError, TypeError):
```
ca-scribner, line 472: What do we need this for?  Can we add a comment to describe the errors we're suppressing?  This feels like something we should clearly warn future developers about

charms/istio-pilot/src/charm.py
```diff
@@ -441,16 +444,38 @@ def _check_leader(self):
...
+            name, _, _ = socket.gethostbyaddr(target)
+            if _is_hostname(name) and not name.endswith(".svc.cluster.local"):
```
ca-scribner, line 474: `_is_hostname(name)` can never be `True` here, because 5 lines above this we have `if is_hostname(target): ... return`

charms/istio-pilot/src/charm.py
```diff
@@ -441,16 +444,38 @@ def _check_leader(self):
...
+            self.log.info("Could not retrieve the gateway service for using in the CSR.")
             return None
-        gateway_address = _get_gateway_address_from_svc(svc)
-        if gateway_address:
-            return gateway_address
+        return _get_gateway_address_from_svc(svc)
+
+    @property
+    def _cert_subject(self) -> Optional[str]:
+        """Return the certificate subject to be used in the CSR."""
```
ca-scribner, line 463: Can we expand this docstring to cover all the cases?  I'm having a hard time deciding what could or should happen here depending on whether the service has an IP, the service has a hostname, the domain name is specified in config, etc

ca-scribner (commented):
Added a few suggestions but mostly lgtm

I'm getting stuck a little trying to review this because there's so many different combinations.  I think it would help if the docstrings on ingress_gateway_host and _cert_subject described the possible outcomes

charms/istio-pilot/config.yaml
```diff
@@ -17,6 +17,11 @@ options:
     type: string
     default: istio-gateway
     description: Name to use as a default gateway
+  domain-name:
```
ca-scribner, line 20: I'm trying to understand when this is needed, and also I am very networking ignorant :)

Sometimes we've seen LBs give the ingress service an external "ip" that is really a hostname.  In those cases, this setting is not needed as we can get the hostname straight from the service (?).

But iiuc, in some cases this is not true and the LB will provide an external IP even though the user will reach it by some hostname.  And in those cases, we need a way to tell istio-pilot (the charm responsible for talking with the cert provider) this hostname because it not automatically discoverable?

charms/istio-pilot/src/charm.py
ca-scribner: yeah I had the same comment about testing.  I'm mentally trying to decide if the different combinations of ip, domain name, host, (or none) all work.

jnsgruk:
Can we line up the name of the config option with that of Traefik etc? (I think they use external_hostname or something like that?)

charms/istio-pilot/src/charm.py
```diff
@@ -441,16 +444,38 @@ def _check_leader(self):
...
+        target = self._ingress_gateway_host
+
+        # Check if the target is a hostname and return if it is
+        if _is_hostname(target):
+            assert isinstance(target, str), target
+            return target
+
+        # If target is an IP address, try to get the hostname and return it
+        with contextlib.suppress(OSError, TypeError):
+            name, _, _ = socket.gethostbyaddr(target)
```
ca-scribner, line 473: if I understand correctly, we get here when:
* domain-name is not set
* target (the gateway service's address) is an IP address

and then we try to lookup a hostname associated with that IP address.  Is that right?

If yes:
1. does this always work?  If it does, do we even need to let users specify domain-name manually, or can they rely on this alone?
2. if this only works sometimes, is it worth implementing at all?  Or should we just say "if you don't specify a domain name and we can't see it from the load_balancer, then you can't get a cert"?  

If (1) is not true, I suggest considering (2) to simplify the logic.  The "look up a host based on the IP address" behaviour feels a lot harder to predict edge cases for, and it also makes the code more complex to read.  If it really does add value then maybe its worth it, but I'm not sure whether it does.

DnPlas:
> Can we line up the name of the config option with that of Traefik etc? (I think they use external_hostname or something like that?)

Hey @jnsgruk the reason why I did not want to use the same name as them is because `traefik` actually uses the external hostname to configure the traefik provider in a more explicit way, while in istio it is done more implicitly (by the LB configuration, for instance). I did not want users to assume this option was the same as we are limiting it to just use it for TLS configuration purposes.

I could see us having a more explicit configuration for the external hostname (the name users will reach on their browsers), but that seems out of scope of this PR. wdyt?

jnsgruk:
> EDIT: or we could not have the option at all and use our "automatic" discovery and keep it implicit in istio. I will explore this option a bit more.

This sounds like a better plan to me :)

DnPlas:
> > EDIT: or we could not have the option at all and use our "automatic" discovery and keep it implicit in istio. I will explore this option a bit more.
> 
> This sounds like a better plan to me :)

Yeah, definitely. Will update the PR shortly.

charms/istio-pilot/config.yaml
DnPlas: I thought that, but then realised that our charm could actually discover hostnames and IP addresses from the `istio-ingressgateway-workload` assuming that the DNS resolves the hostname to the LB Service address, or in all other cases we'll try with the IP address.

I will update this PR removing the configuration option and letting the charm automatically discover this address.

charms/istio-pilot/src/charm.py
DnPlas: I think I will rename it to `is_ipaddress`, it will be more clear that way as we only are checking that.

charms/istio-pilot/src/charm.py
DnPlas: I have changed the code slightly and added test coverage to the `cert_subject` property.

charms/istio-pilot/src/charm.py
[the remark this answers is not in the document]
DnPlas: I can work on that.

charms/istio-pilot/src/charm.py
DnPlas: The reason why we need this is that often times the ingress gateway `Service` will point to an IP address, it rarely has an actual hostname, this can happen if users set their DNS server to make a domain name resolve to the Service IP address.

In a next iteration of this PR I will remove the configuration option and instead let the charm discover and construct this domain name using all this logic. Let me know what you think and if it is clearer after the update.

charms/istio-pilot/src/charm.py
```diff
@@ -441,16 +444,38 @@ def _check_leader(self):
...
+        # If the configuration option is set, we use it
+        if domain_name := self.model.config["domain-name"]:
+            return domain_name
+
+        # We try getting the ingress gateway IP address in the absence of domain name
         try:
             svc = self._get_gateway_service()
         except ApiError:
-            self.log.info("Could not retrieve the gateway service for configuring the CSR.")
+            self.log.info("Could not retrieve the gateway service for using in the CSR.")
```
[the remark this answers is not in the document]
DnPlas: I think this doesn't apply anymore, let me know

charms/istio-pilot/src/charm.py
```diff
@@ -444,34 +444,37 @@ def _check_leader(self):
...
+    def _cert_subject(self) -> Optional[str]:
+        """Return the certificate subject to be used in the CSR.
+
+        This property takes advantage of the automatic address discovery of the
+        ingress gateway address and uses it to return a name that can be used as
+        a certificate subject when integrated with a TLS certificate provider.
+        The logic is as follows:
+        * Return None if the ingress gateway service address cannot be retrieved
+        * Return a hostname if the address is not an IP (e.g. user-example.com)
+        * Return a hostname if the address is an IP, but there is a hostname that resolves to it
```
kimwnasptd, line 456: @DnPlas this is the piece that is hazy to me. How can you detect if a hostname resolves to this IP?

kimwnasptd: But how is this done? How can it know any DNS hostname that could be pointing to this IP?

ca-scribner: Yeah @kimwnasptd, I had similar questions when @DnPlas and I talked offline today.  I don't have something to point to and say "this is obviously wrong", but this is an area I'm not experienced with so I am not confident in ruling out issues either.

My main worry is this: will `socket.gethostbyaddr()` **always** work, including:
* behind a restrictive proxy
* in an airgapped environment (maybe this matters less - not sure we'd need a CSR at all in an airgapped install?)
My guess is there's probably a use case out there that can break this approach, but I don't know it off hand.  

My secondary worry is whether this might block something even in cases when we don't need a CSR at all.  For example, I'm pretty sure right now if I ran this charm in airgapped the charm will try to execute `socket.gethostbyaddr(LOADBALANCER_IP)` which will raise an exception because it cannot resolve a host.  We can fix that, but my worry is that *we don't have tests right now that catch errors like that*.  So to implement this safely, we need to think that stuff through as well

ca-scribner: Yeah just confirmed this...

Locally, I can do:
```python
import socket
socket.gethostbyaddr('8.8.8.8')
# ('dns.google', [], ['8.8.8.8'])
```

whereas from a lxc container without a network interface, I see:

```
socket.gethostbyaddr('8.8.8.8')
# Traceback (most recent call last):
#   File "<stdin>", line 1, in <module>
# socket.herror: [Errno 2] Host name lookup failure
```

And this code gets called on every `__init__` of the charm, regardless of whether we actually need a CSR

Surely we can work around this, I'm just more worried about the edge cases we haven't thought about yet.

ca-scribner: fwiw, I like controlling this via config rather than automation because:
1. I don't have a sense of where else this might fail
2. this is not a setting I expect people to constantly be getting frustrated by

Specifically for (2), casual users wont need this (they likely don't need a public host and bulletproof certs), and enterprise users wont keep redeploying kubeflow so they'll set this once and move on.  And in a rare case of needing to deploy with this setting a lot, its also controllable from a bundle.yaml file.  

tl;dr: it would be nice to have, but this is a tiny optimization.  the risk vs reward doesn't feel right to me.

DnPlas: I can go back to my original implementation of receiving a configuration option, that is not a problem, that way we could just avoid the edge cases.

DnPlas:
Thanks @ca-scribner for the review, this is ready for another pass.

ca-scribner (approved):
lgtm!

There's CI failures atm which I think are unrelated to this feature, but will block publishing after this merges

PR #386, opened 2024-03-01
feat: add `csr-domain-name` config option  (#381)

DnPlas:
feat: enable csr-domain-name config option so istio-pilot can use it on CSRs

The istio-pilot charm already has a mechanism in place to discover the ingress gateway address from the `Service`, but it is limited to only returning IP addresses, which not all TLS certificate providers accept as a valid cert subject. Having the domain-name config option will allow users to specify the domain name they'd like to use when integrating with TLS certificate operators. This feature expands the support for integrating with TLS certificate providers that cannot issue signed certificates on a CSR that only contains an IP address (like we used to do). This commit also adds some test coverage to test the recently added code.

Fixes #379

NOTE: CI may fail because of #385 and #384 

#### Testing instructions
Assuming you've got a microk8s deployment with 
Ensure the upgrade is done correctly:
1. Deploy `istio-operators`

```
juju deploy istio-gateway istio-ingressgateway --config kind=ingress --channel 1.17/stable --trust
juju deploy istio-pilot --channel 1.17/stable --trust
juju relate istio-pilot istio-ingressgateway
```
2. Wait for them to be active and idle
3. Refresh `istio-pilot` to the version in this PR `juju refresh istio-pilot --channel latest/edge/pr-386 --revision <check-revision-in-CI>`
4. Ensure no errors are logged

Ensure the certificate has the correct domain name
1. Deploy `juju deploy self-signed-certificates`
2. Relate `juju relate self-signed-certificates istio-pilot`
3. Change CSR domain name `juju config istio-pilot csr-domain-name="istio-test.com"`
4. Check the certificate subject, it should be the same as the `csr-domain-name` that was set in (3)
```
# Get the issues certificate, look for it in the unit
$ juju show-unit istio-pilot/0
...
  - relation-id: 0
    endpoint: peers
    related-endpoint: peers
    application-data: {}
    local-unit:
      in-scope: true
      data:
        certificate: |-
          -----BEGIN CERTIFICATE-----
         ...
          -----END CERTIFICATE-----

# Check the subject name
$ openssl x509 -noout -subject -in ca.cert <--- you can save the cert in a file and pass it here
subject=CN = istio-test.com, x500UniqueIdentifier = abfd61dc-629d-4eaa-9b19-9613908e0313

```

PR #387, opened 2024-03-01
ci: use latest/candidate for charmcraft channel

DnPlas:
This commit pins charmcraft to latest/candidate to avoid issues with the least stable edge channel.

This PR aims at fixing the following error:

```
charmcraft pack -v
Starting charmcraft, version 2.5.5.post96+g26e3f02                                                                                 
Logging execution to '/home/dplascen/.local/state/charmcraft/log/charmcraft-20240301-124953.845091.log'                            
Bad charmcraft.yaml content:
- ensure this value has at most 78 characters (in field 'summary')                                    
Full execution log: '/home/dplascen/.local/state/charmcraft/log/charmcraft-20240301-124953.845091.log'     
```
Which was caught in the CI of https://github.com/canonical/istio-operators/pull/386.

It would be better to pin this channel than to adapt an older version of this charm. Open to discussions.

### DIFF
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -63,7 +63,7 @@ jobs:
           channel: ${{ matrix.microk8s-versions }}
           microk8s-addons: "dns hostpath-storage rbac metallb:10.64.140.43-10.64.140.49"
           juju-channel: 3.1/stable
-          charmcraft-channel: latest/edge
+          charmcraft-channel: latest/candidate
   
       - name: Run integration tests
         run: |
@@ -130,7 +130,7 @@ jobs:
           channel: ${{ matrix.microk8s-versions }}
           microk8s-addons: "dns hostpath-storage rbac metallb:10.64.140.43-10.64.140.49"
           juju-channel: 3.1/stable
-          charmcraft-channel: latest/edge
+          charmcraft-channel: latest/candidate
  
       - name: Run observability integration tests
         run: |


ca-scribner:
To fix this, should we instead shorten our summary field?  

iiuc `edge` adds this restriction.  switching to `candidate` will let CI pass today but sometime in a few weeks when this feature moves to `candidate` it will just fail again

DnPlas:
> To fix this, should we instead shorten our summary field?
> 
> iiuc `edge` adds this restriction. switching to `candidate` will let CI pass today but sometime in a few weeks when this feature moves to `candidate` it will just fail again

Yeah @ca-scribner , that was my first thought, but our `charmcraft.yaml` does not have a summary field (:

I was also thinking that maybe for track branches we should pin to certain versions of charmcraft. I agree that this change will land in latest/candidate soon, so I was wondering if we should go that route instead. wdty?

ca-scribner:
oh that is a fun nuance... That sounds like a bug we should probably report then too.  I wonder if the `summary` field, if undefined, gets populated by another field?

I'm +10000 on pinning everything when we put something to a track (microk8s, charmcraft, charm dependencies, ... ).  The track is a release, and a release should be a static picture of what we released, not something with unpinned dependencies.  It should be possible, but I think we'd have to change some of the actions (to accept `revision`)

It also might be a pain in the butt to administer if we don't automate the track cutting process.  For a single case like this its cool, but for a full bundle release it would be a nightmare

DnPlas:
> oh that is a fun nuance... That sounds like a bug we should probably report then too. I wonder if the `summary` field, if undefined, gets populated by another field?

Absolutely agree. I'll file an issue (:

> 
> I'm +10000 on pinning everything when we put something to a track (microk8s, charmcraft, charm dependencies, ... ). The track is a release, and a release should be a static picture of what we released, not something with unpinned dependencies. It should be possible, but I think we'd have to change some of the actions (to accept `revision`)
> 
> It also might be a pain in the butt to administer if we don't automate the track cutting process. For a single case like this its cool, but for a full bundle release it would be a nightmare

I also think the track branches should have a static version, do you think we can pin a charmcraft version here and then we can have a tracking issue for changing this everywhere?

ca-scribner:
Is that possible for our CI?  I haven't tried, but I think everything accepts `channel` as an input rather than revision.  My expectation is we'd have to add revision as a separate argument (and to the actions like [actions-operator](https://github.com/canonical/istio-operators/blob/a9fdf4e711d05d141cc660cccc3808b17d94eeb5/.github/workflows/integrate.yaml#L60) and [upload-charm](https://github.com/canonical/charming-actions/blob/main/upload-charm/action.yaml#L30))

If there's a way to pin it here though, I'm +1

DnPlas:
> Is that possible for our CI? I haven't tried, but I think everything accepts `channel` as an input rather than revision. My expectation is we'd have to add revision as a separate argument (and to the actions like [actions-operator](https://github.com/canonical/istio-operators/blob/a9fdf4e711d05d141cc660cccc3808b17d94eeb5/.github/workflows/integrate.yaml#L60) and [upload-charm](https://github.com/canonical/charming-actions/blob/main/upload-charm/action.yaml#L30))
> 
> If there's a way to pin it here though, I'm +1

I don't think at the moment we can accept revisions or versions, I guess what I meant with my previous comment is that we should pin to a channel that was working back when we published this charm. 

I was reviewing the channels, though, and found out that the `latest/stable` version is 2.5.5, which is being worked ATM and may suffer changes, the same changes we are trying to avoid. An earlier version would be 2.2.0, which was last published on Jan 2023, but that is way before the time we published this charm.

For the time being, perhaps we should pin to `latest/stable` until https://github.com/canonical/charmcraft/issues/1568 gets fixed?

DnPlas:
I have pushed https://github.com/canonical/istio-operators/pull/389, which adds a workaround for the summary issue. Will close this PR.

PR #388, opened 2024-03-04
tests: pin kubeflow-volumes to its 1.8/stable channel

DnPlas:
This charm has been refactored into a sidecar which now requires to be trusted, which makes the CI to fail because it cannot deploy this test dependency. On the other hand, the integration test execution should be kept exactly the same as when v1.17 of the istio-pilot was released, thus pinning the charm dependency is a way to keep things consistent and avoid issues.


NohaIhab (approved):
lgtm

PR #394, opened 2024-03-18
feat: add action to handle SSL values as secrets for TLS configuration

### LINKED ISSUE #380: `ssl-key` and `ssl-crt` removal broke compatibility with a specific TLS use case
### Bug Description

Changes introduced by https://github.com/canonical/istio-operators/pull/338 removed the support for enabling TLS for users that only have the `ssl-key` and `ssl-crt` files generated by an external Certificate Authority (CA). This change affects `istio-operators` v1.17 and latest/edge.

That PR removed configuration options that allowed users to pass these values with the assumption that the integration with `tls-certificates-interface` would be a replacement for it. This assumption considered all cases where users wanted to have a charm to provide certificates (a certificates provider) and handle all the required automation, and even considered the case where users would have an in-house CA so they could use charms like the [`manual-tls-certificates-operator`](https://github.com/canonical/manual-tls-certificates-operator). The only missing scenario is the one described in the first paragraph.

This scenario is still relevant for some users because they can only get the `ssl-key` and `ssl-crt` from a CA.

We should consider options to support this, ideally in a way that can guarantee a good level of security and follows the best recommendations from the maintainers of the tls-certificates interface.

DnPlas:
⚠️ WARNING: This feature has been added due to #380, but will be supported only in 1.17 and 1.21 (and the versions released in between, iff any), after that, it will be removed in favour of integrating with a TLS certificates provider. Documentation will be provided for upgrades and migrations.
---
This commits introduces actions that allow users to configure the TLS 
ingress gateway for a single host directly passing the SSL cert and key 
to the charm.
- save-tls-secret: allows users to pass the ssl-key and ssl-crt values,
  which the charm saves in a juju secret (owned by the charm) and uses
  them to reconcile the ingress Gateway with such information.
- remove-tls-secret: a handy action that allows users to remove the 
  TLS secret, which in turn removes the TLS configuration from the 
  ingress Gateway.

This commit also adds unit and integration tests to increase the 
coverage due to the recent changes.

WARNING: please note this feature is only supported in 1.17 and 1.18,
and it will be removed after releasing 1.18 in favour of the TLS 
provider method.

Fixes #380

#### Manual testing instructions

This feature requires juju 3.x

1. Deploy istio-operators from `latest/edge` and relate them
2. Pack the `istio-pilot` charm and refresh. Wait for active and idle.
3. Make sure the `gateway` resource is not configured for `TLS`:

```bash
$ kubectl get gateways -n istio-secrets istio-gateway -oyaml
apiVersion: networking.istio.io/v1beta1
kind: Gateway
metadata:
  creationTimestamp: "2024-03-18T13:01:38Z"
  generation: 5
  labels:
    app.juju.is/created-by: istio-pilot
    app.kubernetes.io/instance: istio-pilot-istio-secrets
    kubernetes-resource-handler-scope: gateway
  name: istio-gateway
  namespace: istio-secrets
  resourceVersion: "46114"
  uid: 57374582-c0fd-4b80-96e8-138196371cf9
spec:
  selector:
    istio: ingressgateway
  servers:
  - hosts:
    - '*'
    port:
      name: http
      number: 8080
      protocol: HTTP
```

4. Run the `save-tls-secret` action to pass values (strings)

```bash
$ juju run istio-pilot/0 save-tls-secret ssl-key=foo ssl-crt=bar
Running operation 111 with 1 task
  - task 112 on unit-istio-pilot-0

Waiting for task 112...
```

5. Confirm these values have been passed to the `Secret` the gateway uses for TLS:

```bash
$ kubectl get gateways -n istio-secrets istio-gateway -oyaml
apiVersion: networking.istio.io/v1beta1
kind: Gateway
metadata:
  creationTimestamp: "2024-03-18T13:01:38Z"
  generation: 6
  labels:
    app.juju.is/created-by: istio-pilot
    app.kubernetes.io/instance: istio-pilot-istio-secrets
    kubernetes-resource-handler-scope: gateway
  name: istio-gateway
  namespace: istio-secrets
  resourceVersion: "46454"
  uid: 57374582-c0fd-4b80-96e8-138196371cf9
spec:
  selector:
    istio: ingressgateway
  servers:
  - hosts:
    - '*'
    port:
      name: https
      number: 8443
      protocol: HTTPS # <---- it is now using HTTPS
    tls:
      credentialName: istio-gateway-gateway-secret # <---- It is now pointing at a secret
      mode: SIMPLE
$ kubectl get secret -nistio-secrets istio-gateway-gateway-secret -oyaml
apiVersion: v1
data:
  tls.crt: YmFy # <--- base64 encoded string
  tls.key: Zm9v # <--- base64 encoded string
kind: Secret
metadata:
  creationTimestamp: "2024-03-18T15:24:48Z"
  labels:
    app.juju.is/created-by: istio-pilot
    app.kubernetes.io/instance: istio-pilot-istio-secrets
    kubernetes-resource-handler-scope: gateway
  name: istio-gateway-gateway-secret
  namespace: istio-secrets
  resourceVersion: "46453"
  uid: f77394a2-a773-4407-8c1f-eb6a18db996c
type: kubernetes.io/tls

```

6. Test other cases:
* Remove the secret with `remove-tls-secret` action and watch the Gateway be reconfigured w/o TLS
* Pass only one value to the `save-tls-secret` action and watch the unit go to BlockedStatus
* Do an upgrade from 1.17/stable to this version of the charm


charms/istio-pilot/actions.yaml
NohaIhab, line 1: For the use case at hand, is there any reason a user would want to set with the action only the key or only the cert? If not, we can make both inputs `ssl-key` and `ssl-crt` required to run the action by adding ` required: [ssl-key, ssl-crt]`

NohaIhab (commented):
thank you @DnPlas 
left some comments, and I suggest @ca-scribner gives it a look as well

charms/istio-pilot/actions.yaml
```diff
@@ -0,0 +1,21 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+save-tls-secret:
```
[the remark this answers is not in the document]
DnPlas: I also thought of that, but I did not want this to be confused with configuring TLS certs via a TLS provider, that's why I wanted `save-tls-secret`, but you are right that this is internal to the charm and from the user perspective this is a configuration of TLS, rather.
I'm not sure if `configure-tls-certs` is appropriate though, it could be `configure-tls-manually`(?)

charms/istio-pilot/actions.yaml
```diff
@@ -0,0 +1,21 @@
...
+  description: |
+    Save SSL cert and key values in a juju secret.
+    The charm will use this information to configure the Ingress Gateway with TLS.
+  params:
+    ssl-key:
+      type: string
+      description: |
+        The SSL key output as a string. Can be set with
+        $ juju run set-tls-secret istio-pilot/<unit-number> ssl-key="$(cat KEY_FILE)"
```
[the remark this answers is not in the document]
DnPlas: Will update once we agree on the name, good catch.

charms/istio-pilot/src/charm.py
```diff
@@ -828,7 +894,83 @@ def _istiod_svc(self):
...
+    # ---- FIXME: Remove this block after releasing 1.18.
+    def _use_https(self) -> bool:
+        """Return True if only one of the TLS configurations are enabled, False if none are.
+
+        Raises:
+            ErrorWithStatus: if both configurations are enabled.
+        """
+        if self._use_https_with_tls_provider() and self._use_https_with_tls_secret():
+            raise ErrorWithStatus(
+                "Only one TLS configuration is supported at a time.", BlockedStatus
```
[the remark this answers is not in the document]
DnPlas: Yeah, will add some instructions, but they will have to go in the logs as the tiny message in the status won't show everything.

charms/istio-pilot/actions.yaml
DnPlas: Yes, I have added the required bit, I missed it completely. Thanks!

charms/istio-pilot/tests/unit/test_charm.py
```diff
@@ -1503,6 +1507,164 @@ def test_wait_for_update_rollout_timeout(self, mocked_istioctl):
...
+    def test_use_https(
+        self,
+        https_with_tls_provider,
+        https_with_tls_secret,
+        expected_return,
+        expected_context,
+        harness,
+        mocked_cert_subject,
+    ):
+        """Test the method returns a correct bool when comparing two TLS options."""
```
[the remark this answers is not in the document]
DnPlas: Hey @NohaIhab I have expanded the test docstrings a bit, the method itself is also a good reference for what the expectation is. Please let me know if this suffices.

charms/istio-pilot/actions.yaml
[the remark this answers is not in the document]
NohaIhab: `configure-tls-manually` works for me to differentiate between the relations way of configuring

charms/istio-pilot/src/charm.py
```diff
@@ -828,7 +894,97 @@ def _istiod_svc(self):
...
+            return False
+
+        # Fail if ssl is only partly configured as this is probably a mistake
+        ssl_key = secret.get_content()["ssl-key"]
+        ssl_crt = secret.get_content()["ssl-crt"]
+        if _xor(ssl_key, ssl_crt):
+            missing = "ssl-key"
+            if not secret.get_content()["ssl-crt"]:
+                missing = "ssl-crt"
+            raise ErrorWithStatus(f"Missing {missing}, cannot configure TLS.", BlockedStatus)
```
NohaIhab, line 971: Is this necessary now that you've added the `required` bit?

charms/istio-pilot/src/charm.py
```diff
@@ -828,7 +894,97 @@ def _istiod_svc(self):
...
+
+        # If both the SSL key and cert are provided, we configure TLS
+        return True if ssl_key and ssl_crt else False
```
NohaIhab, line 974: is this necessary now that you've added the `required` bit?

charms/istio-pilot/src/charm.py
```diff
@@ -828,7 +894,97 @@ def _istiod_svc(self):
...
+
+        Raises:
+            ErrorWithStatus: if both configurations are enabled.
+        """
+        if self._use_https_with_tls_provider() and self._use_https_with_tls_secret():
+            self.log.info(
+                "Only one TLS configuration is supported at a time."
+                "Either remove the TLS certificate provider relation,"
+                "or the TLS manual configuration with the remove-tls-secret action."
+            )
```
NohaIhab, line 943: I think this should be a `warn` (or an `error`?) since we're blocking for this reason

charms/istio-pilot/tests/unit/test_charm.py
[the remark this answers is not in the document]
NohaIhab: lgtm, thank you!

charms/istio-pilot/actions.yaml
```diff
@@ -0,0 +1,22 @@
...
+set-tls-manually:
+  description: |
+    Manually pass SSL cert and key values to configure the Ingress Gateway with TLS.
+    Configuring TLS with this action is mutually exclusive to using TLS certificate providers.
+  params:
+    ssl-key:
+      type: string
+      description: |
+        The SSL key output as a string. Can be set with
+        $ juju run set-tls-manually istio-pilot/<unit-number> ssl-key="$(cat KEY_FILE)"
```
NohaIhab, line 13: let's move the example to the description of the action and change it to set both the key and cert since now both inputs are required

charms/istio-pilot/actions.yaml
```diff
@@ -0,0 +1,22 @@
...
+    ssl-crt:
+      type: string
+      description: |
+        The SSL cert output as a string. Can be set with
+        $ juju run set-tls-manually istio-pilot/<unit-number> ssl-crt="$(cat CERT_FILE)"
```
NohaIhab, line 18: same as above: let's move the example to the description of the action and change it to set both the key and cert since now both inputs are required

charms/istio-pilot/src/charm.py
```diff
@@ -843,8 +1003,7 @@ def _use_https(self):
                 f"Missing {missing}, cannot configure TLS",
                 BlockedStatus,
             )
-        if self._cert_handler.cert and self._cert_handler.key:
-            return True
+        return True if self._cert_handler.cert and self._cert_handler.key else False
```
ca-scribner, line 1006: Is there a reason why this:
* returns `False` if both `cert` and `key` are unset
* goes to blocked if one of `cert` and `key` are unset

?

Having `cert_handler.enabled==True` and `cert`/`key` unset sounds like its still a problem.  In this case, TLS is requested but not configured.  Returning `False` here means the gateway will be set up without auth, which probably isn't what is desired

charms/istio-pilot/src/charm.py
```diff
@@ -828,7 +894,101 @@ def _istiod_svc(self):
...
+        # sets these values as required, but leaving the check in case somewhere in
+        # the code these values are messed up.
+        if _xor(ssl_key, ssl_crt):
+            missing = "ssl-key"
+            if not secret.get_content()["ssl-crt"]:
+                missing = "ssl-crt"
+            raise ErrorWithStatus(f"Missing {missing}, cannot configure TLS.", BlockedStatus)
+
+        # If both the SSL key and cert are provided, we configure TLS
+        return True if ssl_key and ssl_crt else False
```
ca-scribner, line 978: Is there a reason why this:
* returns `False` if both `cert` and `key` are unset
* goes to blocked if one of `cert` and `key` are unset

?

Having the secret set and `cert`/`key` unset sounds like its still a problem.  In this case, TLS is requested but not configured.  Returning False here means the gateway will be set up without auth, which probably isn't what is desired

charms/istio-pilot/src/charm.py
```diff
@@ -112,6 +119,19 @@ def __init__(self, *args):
         # available, revoked, invalidated, or if the certs relation is broken
         self.framework.observe(self._cert_handler.on.cert_changed, self.reconcile)
 
+        # ---- Start of the block
+        # ---- WARNING: this feature is not recommended, but is supported in 1.17 and 1.18.
+        # ---- For details please refer to canonical/istio-operators#380.
+        # ---- FIXME: Remove this block after releasing 1.18.
+        # Save SSL information and reconcile
+        self.framework.observe(self.on.set_tls_manually_action, self.save_tls_secret)
+        self.framework.observe(self.on.secret_changed, self.reconcile_tls_secret)
```
ca-scribner, line 128: Can we use `self.reconcile` here for handling changes in tls secrets instead of adding a `reconcile_tls_secret`?  The rest of this charm does a full reconcile on any meaningful event to avoid us forgetting about some specific interaction between events and I could see it being helpful here as well.  There's also some failure logic in the reconcile function that we'll lose if we avoid it.

This also applies to [this line](https://github.com/canonical/istio-operators/pull/394/files#diff-d3a4e684bd1a3a1822df5ab7605cfce00c97cd7b52f44ed770526c4f4f80eacdR132)

charms/istio-pilot/src/charm.py
```diff
@@ -181,6 +201,56 @@ def _get_image_config(self):
...
+    # ---- Start of the block
+    # ---- WARNING: this feature is not recommended, but is supported in 1.17 and 1.18.
+    # ---- For details please refer to canonical/istio-operators#380.
+    # ---- FIXME: Remove this block after releasing 1.18.
+    def remove_tls_secret(self, event) -> None:
+        """Remove the secret that saves TLS information and reconcile the ingress gateway."""
+        try:
+            secret = self.model.get_secret(label="istio-tls-secret")
+            secret.remove_all_revisions()
+            self._reconcile_gateway()
```
ca-scribner, line 213: Do we need to `._reconcile_gateway()` here, or can we just rely on the [observed secret events](https://github.com/canonical/istio-operators/pull/394/files#diff-d3a4e684bd1a3a1822df5ab7605cfce00c97cd7b52f44ed770526c4f4f80eacdR128) to trigger the correct reconciliation?

ca-scribner (commented):
Added a first pass with initial thoughts.  Will dig deeper next.

The main comment so far is that I think we should avoid running a partial reconciliation of the charm for handling secret events.  We were burned on that in this charm in the past, and moved to doing a full reconcile on all events.  afaict that same logic works well here too

tests/test_bundle_tls_secret.py
```diff
@@ -0,0 +1,97 @@
...
+    group="networking.istio.io",
+    version="v1alpha3",
+    kind="Gateway",
+    plural="gateways",
+)
+
+
+@pytest.fixture(scope="session")
+def lightkube_client() -> lightkube.Client:
+    client = lightkube.Client(field_manager="kserve")
```
ca-scribner, line 21: copy/paste error

```suggestion
    client = lightkube.Client()
```

charms/istio-pilot/tests/unit/test_charm.py
```diff
@@ -1503,6 +1507,192 @@ def test_wait_for_update_rollout_timeout(self, mocked_istioctl):
...
+        https_with_tls_secret,
+        expected_return,
+        expected_context,
+        harness,
+        mocked_cert_subject,
+    ):
+        """Test the method returns a correct bool when comparing two TLS options.
+
+
+        Parameters:
```
ca-scribner, line 1533:
```suggestion
        """Test the method returns a correct bool when comparing two TLS options.

        Parameters:
```

charms/istio-pilot/actions.yaml
```diff
@@ -0,0 +1,22 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+set-tls-manually:
```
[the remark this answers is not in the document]
DnPlas: I want this to not be confused with setting TLS with a TLS cert provider, though the workflow is different, and we can provide documentation, I can see confusion in our users. If you think otherwise, let me know.

charms/istio-pilot/src/charm.py
```diff
@@ -112,6 +119,19 @@ def __init__(self, *args):
         # available, revoked, invalidated, or if the certs relation is broken
         self.framework.observe(self._cert_handler.on.cert_changed, self.reconcile)
 
+        # ---- Start of the block
+        # ---- WARNING: this feature is not recommended, but is supported in 1.17 and 1.18.
```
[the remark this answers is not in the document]
DnPlas: Will update, we plan to support this in 1.17 and whatever version is compatible with Kubeflow 1.9, so 1.21.

charms/istio-pilot/src/charm.py
DnPlas: Will update, there is no reason to return False here.

charms/istio-pilot/src/charm.py
DnPlas: Updated, no reason to return False here, we just want to return True iff both are set.

charms/istio-pilot/src/charm.py
DnPlas: We need it because there is no `secret created` event, so initially when first creating the secret, nothing will be observed.

tests/test_bundle_tls_secret.py
DnPlas: Oooops

charms/istio-pilot/src/charm.py
DnPlas: Yes, the action won't do anything anyway. Thanks for the suggestion!

DnPlas:
Unit tests are blocked because of #395. The issue should be resolved after merging #396 and rebasing this branch.

charms/istio-pilot/actions.yaml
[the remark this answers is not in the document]
ca-scribner: my vote is calling it just `set-tls` and explaining the nuances of it in the `documentation` field of the action.  Not that it is the only consideration, but I can't think of examples elsewhere that feel similar to including the `-manually` on this command.  But its up to you, I'm not blocking the PR on this

charms/istio-pilot/src/charm.py
ca-scribner: Yeah that's fine with me.  Especially where this is a short-lived feature anyway so not likely someone changes the action by mistake in future

charms/istio-pilot/src/charm.py
```diff
@@ -181,6 +201,56 @@ def _get_image_config(self):
...
+            )
+            return
+
+        # Block the unit if there is a missing parameter, we need both to continue
+        if _xor(ssl_key, ssl_crt):
+            missing = "ssl-key"
+            if not ssl_crt:
+                missing = "ssl-crt"
+            self._log_and_set_status(BlockedStatus(f"Missing {missing}, cannot configure TLS."))
+            return
```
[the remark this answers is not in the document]
ca-scribner: This no longer matters because using the secret-based certs is now handled fully by the reconcile function

charms/istio-pilot/src/charm.py
DnPlas: It's updated, please let me know what you think (:

charms/istio-pilot/actions.yaml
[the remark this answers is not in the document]
DnPlas: It's fine, let's go with `set-tls` and have it very clear in the docs.

ca-scribner (approved):

Issue #398, opened 2024-04-02
Use secrets configuration instead of wrapping the secret in action


DnPlas:
### Context

#394 and #397 introduced an action to pass TLS certs and keys to `istio-pilot` to save them as charm secrets and configure the Ingress Gateway with TLS. On https://github.com/canonical/istio-operators/issues/380#issuecomment-2020861148 it is suggested to replace the action with a configuration option and use user secrets instead.

Please note that for this refactor, the desired juju version to use is 3.4, as juju 3.1 does not support user supplied secrets. Please check https://github.com/canonical/bundle-kubeflow/issues/859 for more details on the migration.

### What needs to get done

Replace the logic in the charm that observes the action and use a configuration option to pass user secret IDs. This requires us to:
1. Remove the `set-tls` and `unset-tls` actions
2. Add a configuration option to save user secret IDs
3. Use the secret ID to retrieve the contents of the secret
4. Write some unit tests for this

This change has to be done both in `main` and `track/1.17`

### Definition of Done

The `set-tls` and `unset-tls` actions are removed and replaced with a configuration option to pass user secrets.

DnPlas:
An initial approach at solving this issue has been made in #401. Please note some test cases in that PR are skipped because of https://github.com/canonical/operator/issues/1175.

PR #400, opened 2024-04-02
refactor: use secret config instead of action

DnPlas:
The recommended way of handling user secrets is by saving the user secret ID in a charm configuration, that way the charm can use it to get the contents of the secret. When the charm is granted with the user secret via the CLI, it allows the charm to observe secret events, which help handle changes in their contents. The istio-pilot charm implemented a way to use actions to wrap the secret logic, but because of the above, this commit does a small refactor to remove the set-tls and unset-tls actions and instead directly use user secrets.

This commit also renames ssl to tls to be more accurate when referring to the actual configuration that gets done in the Gateway.

Part of #398


DnPlas:
Closing as I'll be working on #401 first.

PR #401, opened 2024-04-02
refactor: use secret config instead of action for configuring TLS ingress gateway

DnPlas:
⚠️ WARNING: This feature has been added due to #380, but will be supported only in 1.17 and 1.22 (and the versions released in between, iff any), after that, it will be removed in favour of integrating with a TLS certificates provider. Documentation will be provided for upgrades and migrations.
---
The recommended way of handling user secrets is by saving the user secret ID in a charm configuration, that way the charm can use it to get the contents of the secret. When the charm is granted with the user secret via the CLI, it allows the charm to observe secret events, which help handle changes in their contents. The istio-pilot charm implemented a way to use actions to wrap the secret logic, but because of the above, this commit does a small refactor to remove the set-tls and unset-tls actions and instead use user secrets directly.

This commit also renames ssl to tls to be more accurate when referring to the actual configuration that gets done in the Gateway.

Part of #398

Merge after #402 

#### Manual testing

NOTE: this feature works with juju > 3.3, installing and bootstrapping a controller version 3.4 is recommended to be in sync with the CI (changes introduced by #402).

* Add a secret to configure Ingress Gateway with TLS

1. Build and deploy the charms from this branch. Relate them and wait for them to become active and idle.
2. Add a user secret `juju add-secret istio-tls-secret tls-crt=<some str> tls-key=<another str>`
3. Grant access for `istio-pilot` to access the secret contents `juju grant-secret istio-tls-secret istio-pilot`
4. From the output of command in step 2, pass the secret ID to the `istio-pilot` charm as a config option with `juju config istio-pilot tls-secret-id=secret:<secret ID resulting from step 2>`. If you have lost it, you can always check with `juju secrets`.
5. Wait for the charm to become active and idle
6. Verify the `Gateway` resource is configured with TLS and that the `Secret` (the kubernetes resource) has the expected values:

```
$ kubectl get gateways -n istio-secrets istio-gateway -oyaml
apiVersion: networking.istio.io/v1beta1
kind: Gateway
metadata:
  creationTimestamp: "2024-03-18T13:01:38Z"
  generation: 6
  labels:
    app.juju.is/created-by: istio-pilot
    app.kubernetes.io/instance: istio-pilot-istio-secrets
    kubernetes-resource-handler-scope: gateway
  name: istio-gateway
  namespace: istio-secrets
  resourceVersion: "46454"
  uid: 57374582-c0fd-4b80-96e8-138196371cf9
spec:
  selector:
    istio: ingressgateway
  servers:
  - hosts:
    - '*'
    port:
      name: https
      number: 8443
      protocol: HTTPS # <---- it is now using HTTPS
    tls:
      credentialName: istio-gateway-gateway-secret # <---- It is now pointing at a secret
      mode: SIMPLE
$ kubectl get secret -nistio-secrets istio-gateway-gateway-secret -oyaml
apiVersion: v1
data:
  tls.crt: YmFy # <--- base64 encoded string
  tls.key: Zm9v # <--- base64 encoded string
kind: Secret
metadata:
  creationTimestamp: "2024-03-18T15:24:48Z"
  labels:
    app.juju.is/created-by: istio-pilot
    app.kubernetes.io/instance: istio-pilot-istio-secrets
    kubernetes-resource-handler-scope: gateway
  name: istio-gateway-gateway-secret
  namespace: istio-secrets
  resourceVersion: "46453"
  uid: f77394a2-a773-4407-8c1f-eb6a18db996c
type: kubernetes.io/tls
```

* Check the TLS configuration changes if the secret changes

1. Change the secret contents `juju update-secret secret:<secret id> tls-crt=new-crt tls-key=new-key`
2. Check a `secret-changed`` message in `juju debug-log`. The `juju status` msg should mention that the TLS configuration is being applied.
3. Wait for the charm to become active and idle
4. Verify the kubernetes `Secret` to confirm the new values are there

* Remove the TLS configuration

1. Remove the secret ID from the `istio-pilot` configuration `juju config istio-pilot tls-secret-id=""`
2. Remove the user secret from the model `juju remove-secret istio-tls-secret`
3. Observe the `Gateway` is not configured with TLS and the kubernetes `Secret` does not exist anymore

* Verify that the charm can be upgraded from 1.17/stable to the charm in this branch

### TODO
- [x] Refactor integration tests


charms/istio-pilot/src/charm.py
ca-scribner, line 1: (sorry, this is for `metadata.yaml` but I can't comment on that file since it hasn't changed yet)

Since we now need Juju>=3.4, can we add an [assumes block](https://juju.is/docs/sdk/metadata-yaml#heading--assumes) in the metadata.yaml?  I think this would work and block users from deploying to earlier juju verisons:
```yaml
assumes:
- juju >= 3.4
```

charms/istio-pilot/config.yaml
```diff
@@ -24,3 +24,6 @@ options:
     type: string
     default: istio-ingressgateway-workload
     description: Name of the service created by istio-gateway to use as a Gateway
+  tls-secret-id:
+    type: secret
+    description: A configuration option to store the user secret ID that stores the TLS certificate and key values.
```
ca-scribner, line 29: Can we add to this description what the secret should contain?  The code looks for content in the `tls-crt` and `tls-key` keys, so we should explain that here.

charms/istio-pilot/src/charm.py
```diff
@@ -124,13 +125,19 @@ def __init__(self, *args):
...
-        self.framework.observe(self.on.secret_remove, self.reconcile)
+        # Remove TLS information and reconcile
+        # NOTE: contrary to what users would think, removing a user secret
+        # via the CLI "juju remove-secret <secret:secret-ID> will not trigger
+        # a secret_remove event.
+        # The way I have found a workaround is to remove the config option
+        # that stores the secret ID, that way the logic for handling a config-changed
+        # runs and updates the bits where the secret contents are used, in our case
+        # the TLS configuration of the Gateway.
+        # self.framework.observe(self.on.secret_remove, self.reconcile)
```
ca-scribner, line 140: I wonder about keeping `on.secret_remove` here even thought it doesn't work as we expect atm.  I could see the current behaviour being a bug and it being fixed in a future juju version

If we do `.observe(secret_remove)`, I can't think of what we might break by mistake.  So I think its low risk to uncomment this.  wdyt?
```suggestion
        self.framework.observe(self.on.secret_remove, self.reconcile)
```

charms/istio-pilot/src/charm.py
```diff
@@ -878,47 +852,63 @@ def _use_https(self) -> bool:
...
+                f"Also make sure the name of the secret is {TLS_SECRET_LABEL}",
+                BlockedStatus,
+            )
+        except ModelError as e:
+            if "ERROR permission denied" in e.args:
+                # Block the unit when there is an ERROR permission denied
+                raise ErrorWithStatus(
+                    "Access to the istio-tls-secret must be granted.",
+                    BlockedStatus,
+                )
```
ca-scribner, line 886: Can we make this more actionable?  Maybe suggest to see `juju grant-secret --help` for more details?  I'm thinking of helping a user that doesn't really know about how secrets should work.  

"Permission denied trying to access `istio-tls-secret` (id: <secret-id>).  See `juju grant-secret --help` for details on granting permission"?  Its long though.  Could also write something short here telling them to see the logs for more instruction

charms/istio-pilot/src/charm.py
```diff
@@ -836,21 +810,21 @@ def _ssl_info(self) -> Dict[str, str]:
         # FIXME: remove the try/catch block and just return the dictionary that contains
         # data from the CertHandler after 1.21
         try:
-            ssl_secret = self.model.get_secret(label=TLS_SECRET_LABEL)
+            tls_secret = self.model.get_secret(id=self._tls_secret_id, label=TLS_SECRET_LABEL)
```
ca-scribner, line 813: I'm asking Juju as well, but do you see a reason why the `label` helps us?  afaict, we never know if this is the first time we're accessing the secret, so we never know if we need to set the label.  To me I think we should just remove the label entirely since we never access the secret by label.  wdyt?

charms/istio-pilot/src/charm.py
```diff
@@ -878,47 +852,63 @@ def _use_https(self) -> bool:
...
 
-        # Ensure the secret that holds the values exist otherwise fail as this
-        # is an error in our code
+        # Check if the tls-secret-id holds any value and get the secret
+        if not self._tls_secret_id:
+            return False
+
         try:
-            secret = self.model.get_secret(label=TLS_SECRET_LABEL)
+            secret = self.model.get_secret(id=self._tls_secret_id, label=TLS_SECRET_LABEL)
```
ca-scribner, line 872: I'm asking Juju as well, but do you see a reason why the label helps us? afaict, we never know if this is the first time we're accessing the secret, so we never know if we need to set the label. To me I think we should just remove the label entirely since we never access the secret by label. wdyt?

ca-scribner (commented):
lgtm apart from a few minor comments

Will do another pass after the blocking PR is merged/rebased

charms/istio-pilot/src/charm.py
DnPlas: Oh, I had it, but never commited it hehe. Will push it!

charms/istio-pilot/src/charm.py
DnPlas: That is true, I was also debating to keep this or not. The only reason to keep it is to prevent users from passing the ID of another secret.

charms/istio-pilot/src/charm.py
DnPlas: I can expand the message (: good suggestion

charms/istio-pilot/config.yaml
DnPlas: This is an awkward one because this configuration option is just for the secret id, not for the secret itself. Since we don't have a `secret.yaml` file or something similar, I think adding that explanation here could be misleading. Do you think a better place to put that info is in the public docs?

charms/istio-pilot/src/charm.py
DnPlas: Updated, please let me know what you think.

charms/istio-pilot/src/charm.py
orfeas-k: Let's see what juju team says, otherwise I 'm fine with both approaches.

charms/istio-pilot/config.yaml
orfeas-k: For that, I think that we **could** add this information here. Since there is no other way to let users know what the contents of the secret should be. Apart from it, I think we should document it also in the charm's [README](https://github.com/canonical/istio-operators/tree/main/charms/istio-pilot).

orfeas-k (commented):
Really nice work @DnPlas! I left some comments

charms/istio-pilot/src/charm.py
DnPlas: Removing it as the docs description is entirely different from the way we expect (this event to be triggered when the secret is removed from the model). If this behaviour is changed in the future, I can add it.

charms/istio-pilot/src/charm.py
```diff
@@ -884,25 +856,20 @@ def _ssl_info(self) -> Dict[str, str]:
 
         # FIXME: remove the try/catch block and just return the dictionary that contains
         # data from the CertHandler after 1.21
```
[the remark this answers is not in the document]
DnPlas: Updated it, it was a previous implementation.

charms/istio-pilot/src/charm.py
```diff
@@ -676,17 +648,17 @@ def _reconcile_gateway(self):
...
         }
 
         # Secure the gateway, if certificates relation is enabled and
         # both the CA cert and key are provided
         if self._use_https():
             self._log_and_set_status(MaintenanceStatus("Setting TLS Ingress"))
-            context["ssl_crt"] = self._ssl_info["ssl-crt"]
-            context["ssl_key"] = self._ssl_info["ssl-key"]
+            context["tls_crt"] = self._tls_info["tls-crt"]
+            context["tls_key"] = self._tls_info["tls-key"]
```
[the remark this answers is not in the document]
DnPlas: Ack, so no changes required here?

charms/istio-pilot/src/charm.py
```diff
@@ -927,47 +894,64 @@ def _use_https(self) -> bool:
...
-            missing = "ssl-key"
-            if not secret.get_content(refresh=True)["ssl-crt"]:
-                missing = "ssl-crt"
-            self.log.error(f"Missing {missing}, this is most likely an error with the charm.")
-            raise GenericCharmRuntimeError(f"Missing {missing}, cannot configure TLS.")
-        elif not ssl_key and not ssl_crt:
+        try:
+            secret = self.model.get_secret(id=self._tls_secret_id)
+        except SecretNotFoundError as secret_error:
+            if "not granted access to" in secret_error.args[0]:
```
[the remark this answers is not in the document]
DnPlas: I have filed https://github.com/canonical/operator/issues/1229 for this because the behaviour between Harness and the actual deployed charm differs. I will change the charm code and test according to the actual behaviour, but I will skip the unit test case, otherwise it will fail.

charms/istio-pilot/src/charm.py
```diff
@@ -927,47 +894,64 @@ def _use_https(self) -> bool:
...
+            )
+
+        try:
+            tls_key = secret.get_content(refresh=True)["tls-key"]
+            tls_crt = secret.get_content(refresh=True)["tls-crt"]
+        except KeyError as ke:
             self.log.error(
-                "Missing both SSL key and cert values, this is most likely an error with the charm."
+                f"Cannot configure TLS - Missing TLS {ke.args} value(s), "
+                "make sure they are passed as contents of the {TLS_SECRET_LABEL} secret."
```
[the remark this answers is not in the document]
DnPlas: Oops, this is from a previous version of this PR, I have removed it. Good catch!

requirements-integration.txt
```diff
@@ -82,7 +82,7 @@ jedi==0.19.0
     # via ipython
 jinja2==3.1.2
     # via pytest-operator
-juju==3.2.2
+juju==3.4.0.0
```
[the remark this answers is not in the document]
DnPlas: There is no python-libjuju 3.5, see [here](https://pypi.org/project/juju/#history) and [here](https://github.com/juju/python-libjuju/releases/tag/3.4.0.0).

charms/istio-pilot/config.yaml
DnPlas: Would it be helpful if I add the secret description in the README and mention the secret contents should passed as mentioned in that doc?

charms/istio-pilot/src/charm.py
DnPlas: Resolving as I have accepted the suggestion

charms/istio-pilot/src/charm.py
DnPlas: I have removed the label entirely, nothing else to do. Resolving this conversation.

charms/istio-pilot/src/charm.py
DnPlas: See https://github.com/canonical/istio-operators/pull/401#discussion_r1550018604, resolving

charms/istio-pilot/config.yaml
orfeas-k: Sure this makes sense!

charms/istio-pilot/src/charm.py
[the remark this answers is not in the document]
orfeas-k: Sure, let's watch this so if they fix it we can update it here as well.

charms/istio-pilot/tests/unit/test_charm.py
```diff
@@ -1601,6 +1601,9 @@ def test_use_https_with_tls_secret_not_found_with_id(
             in error_with_status.value.msg
         )
 
+    @pytest.mark.skip(
+        "Harness is not modeling the granted access correctly, see canonical/operator#1229"
+    )  # noqa
```
orfeas-k, line 1606: Should we add a `FIXME` or `TODO`? I think that it's not clear that once the referenced issue is resolved, we expect to remove the `skip`

DnPlas: I have added a `FIXME` in both the charm and the test file to make it more clear. Let me know what you think.

charms/istio-pilot/config.yaml
DnPlas: Thinking about this more, since the docs are not published yet, I will just add a note in the description of the config option. I will then submit a PR for adding instructions in the README and in public docs. let me know what you think.

orfeas-k (approved):
Good job once more! Could you also include the PR's title/commit the this refers to TLS?

PR #409, opened 2024-04-15
tests: use stable versions of COS and TLS charms

DnPlas:
This commit pins all charms that get deployed from Charmhub to a compatible version to the CKF 1.8 release.

Part of canonical/bundle-kubeflow#863

### DIFF
--- a/tests/test_bundle.py
+++ b/tests/test_bundle.py
@@ -135,7 +135,11 @@ async def test_gateway_info_relation(ops_test: OpsTest):
     TODO (https://github.com/canonical/istio-operators/issues/259): Change this from using a
      specific charm that implements ingress's requirer interface to a generic charm
     """
-    await ops_test.model.deploy(TENSORBOARD_CONTROLLER, channel=TENSORBOARD_CONTROLLER_CHANNEL, trust=TRUST_TENSORBOARD_CONTROLLER)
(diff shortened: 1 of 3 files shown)


misohu (commented):
This PR looks bood but in the test_bundle.py we use inconsistent naming convetion with rest of our test file eg. [here](https://github.com/canonical/istio-operators/blob/1f4822f6eadfebef2a9443868203cef9b19d91d8/tests/test_bundle.py#L25) instead of 

```
TRUST_DEX_AUTH = True
```
we should use 
```
DEX_AUTH_TRUST
```

Same problem is with all the other TRUST variables. Maybe we should change that in this PR ?

DnPlas:
> This PR looks bood but in the test_bundle.py we use inconsistent naming convetion with rest of our test file eg. [here](https://github.com/canonical/istio-operators/blob/1f4822f6eadfebef2a9443868203cef9b19d91d8/tests/test_bundle.py#L25) instead of
> 
> ```
> TRUST_DEX_AUTH = True
> ```
> 
> we should use
> 
> ```
> DEX_AUTH_TRUST
> ```
> 
> Same problem is with all the other TRUST variables. Maybe we should change that in this PR ?

I have updated this PR and sent a fix for `main` in #410

orfeas-k (approved):

PR #438, opened 2024-06-06
fix: remove default nodeAffinity configurations

### LINKED ISSUE #426: `nodeSelectorTerms` gets ignored because of OR logic
### Bug Description

`istio-gateway 1.16` has a default `nodeSelectorTerms` for `requiredDuringSchedulingIgnoredDuringExecution` (see [here](https://github.com/canonical/istio-operators/blob/track/1.16/charms/istio-gateway/src/manifest.yaml#L80) that causes any custom `nodeSelectorTerm` injected by the `namespace-node-affinity-operator` to be "ORed", resulting in an unexpected scheduling as the default could be True before evaluating the injected term. That happens because:

>If you specify multiple terms in nodeSelectorTerms associated with nodeAffinity types, then the Pod can be scheduled onto a node if one of the specified terms can be satisfied (terms are ORed).
[Reference](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/)

Please note this issue happens only in `1.16`, as greater versions will not have a default option (removed by https://github.com/canonical/istio-operators/commit/88c9b987a0371e88e574574c5917e79ca965418e)
### To Reproduce

1. Deploy istio-operators 1.16/stable
2. Deploy namespace-node-affinity
3. Pass a configuration to inject custom nodeSelectorTerms
4. Depending on the custom configuration you will have to  check the Pod is scheduled in the desired Node

### Environment

1. Kubernetes cluster with at least two nodes
2. istio-operators 1.16/stable
3. Nodes are labeled so that we can schedule Pods based on their nodeSelectors

### Relevant Log Output

N/A

### Additional Context

[Similar issue](https://github.com/canonical/namespace-node-affinity-operator/issues/35)

DnPlas:
The default values of the nodeAffinity field were causing this charm to cause undesired behaviours when trying to replace them (either manually or via the namespace-node-affinity charm. This commit removes those defaults from istio-gateway 1.16, newer versions of this charm do not present this issue.

Fixes #426

#### Testing

1. [Create a microk8s cluster](https://microk8s.io/docs/clustering) (2 worker nodes, 1 admin node). You can do it with [LXD containers](https://microk8s.io/docs/install-lxd).
2. Label one of the nodes with an obvious label, it can be `kubectl label node workload=istio-deploys-here`
3. For testing purposes, the admin node will also have a juju controller bootstrapped inside, but this can be done using a controller bootstrapped elsewhere.
4. Add a model `istio-test` and label it with `kubectl label ns testing-ns-a namespace-node-affinity=enabled`, this is a requirement from the `namespace-node-affinity` charm.
5. Deploy namespace-node-affinity `juju deploy namespace-node-affinity --channel 2.2/stable --trust`.
6. Deploy the `istio-operators` from this PR (you can use the [published](https://github.com/canonical/istio-operators/actions/runs/9400421611/job/25890004228#step:5:15) charms)
7. Observe that the `istio-ingressgateway-workload-xxxx-yyyy` container gets scheduled in a random node with no specific criteria.
8. Create a configuration file for passing as configuration to the `namespace-node-affinity`:

```
root@microk8s-admin:~# cat settings.yaml 
istio-test: |
  nodeSelectorTerms:
    - matchExpressions:
      - key: workload
        operator: In
        values:
        - istio-deploys-here
```
9. Pass the configuration file as config to the `namespace-node-affinity`:

```
SETTINGS_YAML=$(cat settings.yaml)
juju config namespace-node-affinity settings_yaml="$SETTINGS_YAML"
```
10. Wait for everything to get active and idle
11. Remove the `istio-ingressgateway-workload-xxxx-yyyy` `Pod` from the namespace and wait for it to be scheduled again.
12. The `Pod` should now be scheduled in the labeled `Node` that was specified in the `nodeSelectorTerms` from step 8.

### DIFF
--- a/charms/istio-gateway/src/manifest.yaml
+++ b/charms/istio-gateway/src/manifest.yaml
@@ -55,37 +55,8 @@ spec:
     spec:
       affinity:
         nodeAffinity:
-          preferredDuringSchedulingIgnoredDuringExecution:
-            - preference:
-                matchExpressions:
-                  - key: kubernetes.io/arch
-                    operator: In
-                    values:
-                      - amd64
-              weight: 2
-            - preference:
-                matchExpressions:
-                  - key: kubernetes.io/arch
-                    operator: In
-                    values:
-                      - ppc64le
-              weight: 2
-            - preference:
-                matchExpressions:
-                  - key: kubernetes.io/arch
-                    operator: In
-                    values:
-                      - s390x
-              weight: 2
-          requiredDuringSchedulingIgnoredDuringExecution:
-            nodeSelectorTerms:
-              - matchExpressions:
-                  - key: kubernetes.io/arch
-                    operator: In
-                    values:
-                      - amd64
-                      - ppc64le
-                      - s390x
+          preferredDuringSchedulingIgnoredDuringExecution: null
+          requiredDuringSchedulingIgnoredDuringExecution: null
       containers:
         - args:
             - proxy
--- a/charms/istio-gateway/tests/unit/data/egress-example.yaml
+++ b/charms/istio-gateway/tests/unit/data/egress-example.yaml
@@ -54,37 +54,8 @@ spec:
     spec:
       affinity:
         nodeAffinity:
-          preferredDuringSchedulingIgnoredDuringExecution:
-            - preference:
-                matchExpressions:
-                  - key: kubernetes.io/arch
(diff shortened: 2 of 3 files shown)


orfeas-k (approved):
Tested and works! Good job @DnPlas!

PR #447, opened 2024-06-17
ci: bump microk8s 1.24(lower) and 1.27(upper)

DnPlas:
Based on Istio's [supported versions](https://istio.io/latest/docs/releases/supported-releases/#support-status-of-istio-releases)
Part of canonical/bundle-kubeflow#903

### DIFF
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -47,8 +47,8 @@ jobs:
     strategy:
       matrix:
         microk8s-versions:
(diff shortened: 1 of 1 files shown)


misohu (commented):
@DnPlas microk8s does not have `1.24-strict/stable`

DnPlas:
> @DnPlas microk8s does not have `1.24-strict/stable`

Let me change that, then, thanks!

misohu (approved):

PR #476, opened 2024-07-19
Fix istio CI

misohu:
Fixes:
- Update jinja2 in requirements.txt https://github.com/canonical/bundle-kubeflow/issues/883#issuecomment-2247353586
- Move TLS dependencies from  charmcraft.yaml to requirtements


DnPlas (approved):
Thanks @misohu !

PR #477, opened 2024-07-19
Add metrics-endpoint to istio-gateway

rgildein:
Add relations endpoints for metrics and Grafana dashboards + update manifest to export metrics via service port. I add separate service port, since `ClusterIP` is enough for metrics.

I also update cos-integration test for both charms using only chisme cos abstraction. For testing if alert rule is firing, we should use promtool.

fixes: #475

PR #479, opened 2024-07-22
Fix charm dependecies

rgildein:
Add cos along with grafana_dashboard and prometheus_scrape, since this is related with #477.

Part of canonical/bundle-kubeflow#989


DnPlas (changes requested):
Please do not merge, as we might need to make a different change. Check https://github.com/canonical/charmcraft/issues/1135#issuecomment-2244878420 for more context.

misohu (commented):
Can we split it into two PRS one for the CI upgrade and one for adding the libs. I understand we want to test it but we can always refer this previous run and split it. Also I miss the information about how did we update the requirements in description.

rgildein:
Closing since @misohu is doing part of it in #476 and second part of bringing cosl to isto-gateway will be done in #477.

Issue #487, opened 2024-07-31
Remove `public-url` configuration from integration tests and add required relation


DnPlas:
### Context

The `oidc-gatekeeper` and `dex-auth` charms do not have this configuration option anymore. It is still set in the integration tests for this repo, so we should remove it and add the appropriate relation so those charms are active and idle and don't block the CI.

### What needs to get done

1. Remove the `public-url` configuration option from integration tests
2. Add the dex-oidc-config relation between oidc and dex charms

### Definition of Done

The CI is not blocked because of the extra config option.

DnPlas:
Track branches from 1.18 -> 1.22 now have the fix. Closing this issue, as there is nothing else to do.

PR #489, opened 2024-07-31
tests: remove public-url config and add dex-oidc-config relation

DnPlas:
The public-url configuration option has been replaced by a relation between dex-auth and oidc-gatekeeper. Because of that, configuring this option is not required anymore, and in fact if kept, can cause errors as it won't be recognised by Juju anymore. This commit removes the extra configuration from the integration tests.

Part of #487


rgildein (approved):
Fixing this in #477, but doing it by separate PR is better :+1:.

DnPlas:
> Fixing this in #477, but doing it by separate PR is better 👍.

Makes sense!

PR #493, opened 2024-07-31
tests: remove public-url config and add dex-oidc-config relation

DnPlas:
The public-url configuration option has been replaced by a relation between dex-auth and oidc-gatekeeper. Because of that, configuring this option is not required anymore, and in fact if kept, can cause errors as it won't be recognised by Juju anymore. This commit removes the extra configuration from the integration tests.

Part of #487


DnPlas:
Re-running due to a transient (known) issue https://github.com/canonical/bundle-kubeflow/issues/1005

DnPlas:
Closing this PR as it is not required. The versions of dex and oidc used for integration tests do not have the `dex-issuer-url` relation. Nothing else to do.

PR #537, opened 2024-08-28
chore(backport): Backport metrics work (KF-5558) to track/1.22

rgildein:
Backporting multiple PR at once, but three of them are adding only dashboard and alert rules. Only the #477 is enabling metrics for istio-gateway. 

This PR needs to me merged with merge commit instead of squash and merge.


DnPlas (approved):
Thanks @rgildein !

PR #568, opened 2024-12-06
chore: replace tls lib v3 and cert handler with tls lib v4

gruyaume:
## Overview

Here we bump the TLS Certificates Library from `v3` to `v4`. This change allows to completely get rid of the `CertHandler` abstraction layer. This change is a courtesy of the Credentials Charm Engineering team.

> :warning: This change is still a draft. I have to properly test it end-to-end prior to making it live.

## Rationale

The TLS Certificates Library `v4` takes care of most of the heavy lifting related to TLS Certificates. This new version of the library includes:
- Automated Private Key Management
- Automated CSR generation
- Automated Certificate management (including renewal)

CertHandler existed purely because of missing feature in TLS library. Now that lib v4 takes care of all of the above, charm authors shouldn't need to import an additional abstraction layer.

## Library Changes
- tls_certificates: v3 -> v4

## Compatibility Notes
This new version of the library still uses the same version of the underlying charm relation interface, meaning that this charm here will remain compatible with the same TLS providers.

## Reference
https://charmhub.io/tls-certificates-interface


gruyaume:
Closing as this effort is not getting prioritised

PR #578, opened 2025-01-17
chore(ci): bump microk8s to 1.29-1.31

orfeas-k:
Bump microk8s 1.29-1.31 according to https://github.com/kubeflow/community/issues/767#issuecomment-2593058929.

Ref canonical/bundle-kubeflow#1157

### DIFF
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -136,8 +136,8 @@ jobs:
     strategy:
       matrix:
         microk8s-versions:
-          - 1.25-strict/stable
-          - 1.26-strict/stable
+          - 1.31-strict/stable
+          - 1.31-strict/stable
     steps:
       - name: Check out repo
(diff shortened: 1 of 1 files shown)


.github/workflows/integrate.yaml
```diff
@@ -56,8 +56,9 @@ jobs:
     strategy:
       matrix:
         microk8s-versions:
-          - 1.27-strict/stable
+          - 1.29-strict/stable
           - 1.30-strict/stable
+          - 1.31-strict/stable
```
NohaIhab, line 61: Can we keep it to 2 versions only?

orfeas-k: yes, we keep the lowest and highest supported versions. On the contrary, we don't have to test o11y with more than one versions. That was implemented by accident.

NohaIhab (approved):

Issue #536, opened 2024-08-27
TLS juju secret removal 


misohu:
### Context

The plan was to remove the TLS juju secret from `istio-pilot` by kubeflow 1.10 (istio operators version > 1.22).
https://github.com/canonical/istio-operators/issues/380
https://github.com/canonical/istio-operators/pull/394

For now we cant remove the secret object because it is the way customers are setting up their TLS termination on ingress level. Customers are generating key and certificate from CA and they set it with the istio-pilot secret objects [`tls-secret-id`](https://github.com/canonical/istio-operators/blob/main/charms/istio-pilot/config.yaml#L41). Currently none of the TLS charms are providing this functionality.
(docs) https://charmhub.io/topics/security-with-x-509-certificates
- https://charmhub.io/route53-lego-k8s
- https://charmhub.io/namecheap-lego-k8s
- https://charmhub.io/httprequest-lego-k8s
- https://charmhub.io/manual-tls-certificates

Another issue we are facing with integration of TLS charms is that the [proposed solution](https://charmhub.io/httprequest-lego-k8s/docs/h-reference-solving-dns-01) requires a server with HTTP endpoint which knows how to talk to DNS server to help resolve the [DNS 01 challenge](https://www.ssl.com/faqs/which-acme-challenge-type-should-i-use-http-01-or-dns-01/) (or any other certificate authority challenge). 

For now we were not considering the implementation of this functionality on istio side, because we need to define if we want all ingress charms to be involved in resolving the challenge, or we want the users to setup the HTTP endpoint of a custom challenge solver for themselves. 

Regardless of the approach we need to expose the end-to-end use case in the charmed kubeflow docs and be able to link external TLS to support this use case (HTTP endpoint challenge solver). 

### What needs to get done

Passing key and CA to charm
- Have a way to pass certificate and key to TLS charms, which will propagate it to requirer charm as explained https://github.com/canonical/istio-operators/issues/380 (sync with telco team about the solution)
- Ensure there are docs 

HTTP endpoint on challenge solver problem 
- Sync with other teams working with ingress and decide if we want the in charm implementation or take it away from charms.
- Ensure there are docs 

Chamred kubeflow
- Create end-to-end doc for charmed kubeflow
- Once we agreed on the design of how the TLS operator charms should handle manual certificates then ensure the istio pilot charm aligns with it (remove the current TLS secret id implementation and add the proposed implementation). 
[...]

PR #592, opened 2025-02-14
(backport) chore: remove deprecation warning for the TLS secret configuration (#590)

DnPlas:
Originally, the TLS configuration that's done by passing a juju secret to istio-pilot was expected to last only a few versions of this charm, but as exposed in #536, this feature is needed until there is a solution that fits out users/customer needs better. Because of that, this commit removes the comments and warnings about deprecating this feature.


orfeas-k (approved):

DnPlas:
CI should be fixed by https://github.com/canonical/istio-operators/pull/601

PR #618, opened 2025-04-29
fix(workaround): set constraints to setuptools-scm to avoid install loops

### LINKED ISSUE #617: Pin binary-packages to fix build failure with Python 3.8
Child issue of https://github.com/canonical/bundle-kubeflow/issues/1255

The charm(s) of this repo all fail to build, due to the issue with Python 3.8 and dependency loops.

To unblock and ensure the charms can build, with Ubuntu 20.04 as base, we'll need to:
1. Remove the top level packages from `requirements.in`
2. Move them, with their specific version, to `charm-binary-python-packages` in `charmcraft.yaml`
3. (Potentially) Update the `requirements-unit.in` to not include `requirements.txt`, but instead any packages that it needs

DnPlas:
Due to python/importlib_metadata#516, we need to set up constaints to the `setuptools-scm` to avoid install loops for these charms dependencies, as recommended by the charmcraft team (see https://github.com/canonical/charmcraft/issues/2259#issuecomment-2842766428).

Fixes #617


charms/istio-pilot/charmcraft.yaml
```diff
@@ -38,30 +38,45 @@ parts:
...
       # Ubuntu 22.04 archive. Install rustc and cargo using rustup instead of the Ubuntu archive
-      rustup set profile minimal
-      rustup default 1.83.0  # renovate: charmcraft-rust-latest
+      #rustup set profile minimal
+      #rustup default 1.83.0  # renovate: charmcraft-rust-latest
+
+      # Install the following packages from binary to avoid python/importlib_metadata#516
+      PIP_BREAK_SYSTEM_PACKAGES=true python3 -m pip install --only-binary :all: charmed-kubeflow-chisme==0.4.6 cosl==0.0.57 \
+        cryptography==44.0.0 jsonschema==4.17.3  lightkube==0.17.1 lightkube-models==1.32.0.8 requests==2.32.3 \
+        serialized-data-interface==0.7.0 tenacity==9.0.0
```
kimwnasptd, line 76: **question**: How come you installed via pip and not `charm-binary-python-packages`? I would expect they are the same, but maybe I'm missing something

DnPlas: This is the python [plugin](https://canonical-charmcraft.readthedocs-hosted.com/en/latest/reference/plugins/python_plugin/), which AFAIK does not have the `charm-binary-python-packages` keyword.

kimwnasptd: aah OK!

